PR #2, opened 2021-06-25
Generate Theme CSS and save to s3 when theme is updated

dandlezzz:
This PR:
Generates a theme css file and saves it to s3 when a theme record is updated.
Saves the URL of theme file to theme record.


lib/updateThemeCss/saveFile.ts
```diff
@@ -0,0 +1,38 @@
...
+  const minioClient = new Minio.Client({
+      endPoint: 's3.amazonaws.com',
+      accessKey: process.env.AWS_ACCESS_KEY,
+      secretKey: process.env.AWS_SECRET_KEY
+  });
+
+
+
+  const hash = SparkMD5.hash(cssContent)
+  const location = `${subdomain}.holaplex.com/${pubkey}/overrides-${hash}.css`
```
kespinola, line 18:
```suggestion
  const location = `${pubkey}/overrides-${hash}.css`
```
The pubkey is all we need.

lib/updateThemeCss/saveFile.ts
```diff
@@ -0,0 +1,38 @@
...
+  const bucketName = 'opus-logica-holaplex-storefronts'
+  const themeUrl = `https://${location}`
+
+  const fs = new Readable()
+  fs.push(cssContent)
+  fs.push(null)
```
kespinola, line 24: This is nice. Never writing to disk.

kespinola (approved):

PR #69, opened 2021-09-30
Remove Storefront Create and Edit Requirement on ArConnect

kespinola:
### Goal

Users no longer need to acquire AR in order to create (or edit) a storefront. Holaplex will pay for all uploads using a treasury account on the server.

https://www.notion.so/holaplex/Remove-ArConnect-on-Storefronts-4e4f5b9d36e443fb8027a57f2937e6b8


pages/api/arweave/upload.ts
```diff
@@ -0,0 +1,138 @@
...
+  type,
+}: {
+  data: Buffer;
+  name: string;
+  type: string;
+}): Promise<ArweaveFile> => {
+  const {
+    arweave,
+    arweaveKeypair: { jwk },
+  } = await WALLETS;
```
kespinola, line 84: I had a PR out that moved accessing arweave through next config. I'll look to get that back should help provide consistent access to arweave throught the app and (like it is now only init once).

pages/api/storefronts.ts
```diff
@@ -1,26 +1,142 @@
...
+import Ajv, { JTDSchemaType } from 'ajv/dist/jtd';
+import { ArweaveFile } from '@/modules/arweave/types';
+import { PutStorefrontParams } from '@/modules/storefront/put-storefront';
+import { unpackNotarized, verifyNaclSelfContained } from '@/modules/notary';
+import { PublicKey } from '@solana/web3.js';
+import { WALLETS } from '@/modules/wallet/server';
+import { stylesheet } from '@/modules/theme';
+
+/** JSON schemas for parsing request parameters. */
+const SCHEMAS = (() => {
```
kespinola, line 15: Schema is another item we could register on the next config. Really anything we are managing with constant closures like this.

src/modules/notary/index.ts
```diff
@@ -0,0 +1,137 @@
+import Ajv, { JTDSchemaType } from 'ajv/dist/jtd';
```
kespinola, line 1: Can you add to the tech spec the high-level operations that are taking place for the message decryption and validation? Help me and new engineers understand it.

src/modules/notary/index.ts
```diff
@@ -0,0 +1,137 @@
...
+  const { payload, signature } = notarized;
+  const parse = options?.parse ?? jsonParse(options.validate);
+  const payloadDec = parse(payload);
+
+  if (payloadDec === undefined) return undefined;
+
+  const msg = await createMessage(payload, options?.format ?? defaultFormat);
+
+  try {
+    if (!(await verify(msg, signature, payloadDec))) return undefined;
```
kespinola, line 106: This is interesting need to create a message in order ot verify?

```suggestion
  const msg = await createMessage(payload, options?.format ?? defaultFormat);

  try {
   cons isVerified = await verify(msg, signature, payloadDec)
    if (!isVerified) return undefined;
```

src/modules/storefront/editor.tsx
```diff
@@ -0,0 +1,227 @@
...
+    const canAfford =
+      arweaveWalletAddress && (await ar.wallet.canAfford(arweaveWalletAddress, file.size));
+
+    if (canAfford) return;
+
+    setShowARModal(true);
+
+    throw new Error('Not enough AR funds to cover the upload fee.');
+  };
+};
```
kespinola, line 163: I'm up for the validate to be unaware of the modal state changes. Can await validateArweaveFunds in the onValidate actions and toggle the setShowARModal within the view. Trying to separate business logic from UI management.

src/modules/storefront/editor.tsx
```diff
@@ -0,0 +1,227 @@
...
+export const validateArweaveFunds = (
+  arweaveWalletAddress: string,
+  ar: ArweaveScope,
+  setShowARModal: (val: boolean) => void
+): ((rule: RuleObject, [file]: [FileInput?]) => Promise<void>) => {
+  return async (_, [file]) => {
+    if (isNil(file) || has('url', file)) return;
+
+    const canAfford =
+      arweaveWalletAddress && (await ar.wallet.canAfford(arweaveWalletAddress, file.size));
```
kespinola, line 155:
```suggestion
    const canAfford = await ar.wallet.canAfford(arweaveWalletAddress, file.size);
```

src/modules/storefront/editor.tsx
```diff
@@ -0,0 +1,227 @@
...
+  };
+};
+
+export const validateArweaveFunds = (
+  arweaveWalletAddress: string,
+  ar: ArweaveScope,
+  setShowARModal: (val: boolean) => void
+): ((rule: RuleObject, [file]: [FileInput?]) => Promise<void>) => {
+  return async (_, [file]) => {
+    if (isNil(file) || has('url', file)) return;
```
kespinola, line 152:
```suggestion
    if (isNil(file) || has('url', file) || arweaveWalletAddress) return;
```

Though arweaveWalletAddress isn't marked as optional so does it existence need to be checked? The type system will catch in potential unknown variable declarations.

src/modules/storefront/editor.tsx
```diff
@@ -0,0 +1,227 @@
...
+      router.push('/').then(() => {
+        track('storefront', trackEvent);
+        setSubmitting(false);
+      });
+    } catch {
+      setSubmitting(false);
+      toast.error(errorToast);
+    }
+  };
+};
```
kespinola, line 224: No need to refactor now but we discussed this function. What I think should still be managed within the view is setting the toast, calling setSumitting, and changing the route. This onSubmit helper should just post the storefront and error.

If the other items are removed can likely consolidate putStorefront into this method as well.

src/modules/storefront/put-storefront.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const notarized: PutStorefrontParams = await notarize(storefront, signPhantom(solana));
+
+    onProgress('uploading');
+
+    const postResp = await fetch('/api/upload-store', {
+      method: 'POST',
+      body: JSON.stringify(notarized),
+    });
+
```
kespinola, line 47:
```suggestion
    const postResp = await fetch('/api/storefronts', {
      method: 'PUT',
      body: JSON.stringify(notarized),
    });
```

How have you incorporated the onProgress? Can you share screenshot?

src/modules/notary/index.ts
ray-kast: Yeah, it's a weird quirk with the API that Phantom exposes - it produces a signature of the message that can be verified but not decrypted, so the only way I've found to verify signatures from Phantom is to recreate the exact message the user would have seen and cross-check it with their pubkey with `tweetnacl`.  There might be a way to sign an embedded message with Phantom but their docs are kinda terse.

src/modules/storefront/editor.tsx
ray-kast: This code actually just needs to go, I plan on removing validateArweaveFunds once I get the new business logic for the Upload component working.

src/modules/storefront/put-storefront.ts
ray-kast: I have not, the WIP commit contains a lot of missing glue code :sweat_smile:

src/modules/arweave/upload.ts
```diff
@@ -0,0 +1,97 @@
...
+    const hash = Buffer.from(await sha256(await file.arrayBuffer(), { outputFormat: 'buffer' }));
+
+    const payload: ArweaveUploadPayload = {
+      fileHash: hash.toString('base64'),
+      pubkey: solana.publicKey.toBase58(),
+    };
+
+    const notarized: ArweaveUploadParams = await notarize(payload, signPhantom(solana), {
+      format: formatMessage,
+    });
```
kespinola, line 59: This this triggers a popup in phantom to sign the transaction.

src/modules/notary/index.ts
```diff
@@ -0,0 +1,168 @@
...
+  sign: (utf8: Buffer) => Promise<Signature>,
+  options?: {
+    format?: (bytes: Buffer) => string;
+  }
+): Promise<Notarized<T>> => {
+  const payloadStr = jsonStringify(payload);
+  const msg = await createMessage(payloadStr, options?.format ?? defaultFormat);
+  const signature = await sign(msg);
+
+  return { payload: payloadStr, signature };
```
kespinola, line 100: Format payload in json. Create message for phantom to sign. Phantom signs message.

src/modules/notary/index.ts
```diff
@@ -0,0 +1,168 @@
...
+export type NotarizedStr<T> = Notarized<T, SignatureStr>;
+
+const defaultFormat = (hash: Buffer) =>
+  `Your transaction fingerprint is: ${hash.toString('base64')}`;
+
+const createMessage = async (payloadStr: string, format: Formatter): Promise<Buffer> => {
+  const payloadBuf = Buffer.from(payloadStr, 'utf-8');
+  const payloadArr = new Uint8Array(
+    payloadBuf.buffer.slice(payloadBuf.byteOffset, payloadBuf.byteOffset + payloadBuf.byteLength)
+  );
```
kespinola, line 63: Pulled from buffer package. Need to slice buffer to unit8 array.

src/modules/notary/index.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export const signNacl =
+  (secret: Uint8Array): Signer =>
+  async (utf8) =>
+    Buffer.from(nacl.sign.detached(utf8, secret)) as Signature;
+
+export const verifyNacl =
+  (pubkey: Uint8Array): Verifier<unknown> =>
+  async (utf8, sig) =>
+    nacl.sign.detached.verify(utf8, sig, pubkey);
```
kespinola, line 82: Allows for verifying when the private key is known. Potential to use on the server.

pages/api/arweave/upload.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  const payloadRes = await resultThenAsync(
+    parseNotarized<ArweaveUploadPayload>(notarized as JsonString<ArweaveUploadParams>),
+    (payload) =>
+      unpackNotarized(
+        payload,
+        verifyNaclSelfContained((s) => new PublicKey(s.pubkey).toBuffer()),
+        { parse: parsePayload, format: formatMessage }
+      )
+  );
```
kespinola, line 60: 1. Parse the unverified payload to get the pubkey that should be the signer.
2. Pass pubkey to verfiy. 
3. Reject or accept based on the signature being from the pub key in the payload.

pages/api/arweave/upload.ts
```diff
@@ -0,0 +1,155 @@
...
+  const file: File = files instanceof Array ? files[0] : files;
+
+  const data = await fs.readFile(file.path);
+  const dataArr = new Uint8Array(
+    data.buffer.slice(data.byteOffset, data.byteOffset + data.byteLength)
+  );
+
+  const hash = Buffer.from(await sha256(dataArr, { outputFormat: 'buffer' }));
+
+  if (hash.toString('base64') !== payload.fileHash) throw new ApiError(400, 'File hash mismatch');
```
kespinola, line 76: Confirm failed provided is the the one signed by the user.

PR #72, opened 2021-10-08
Ipfs upload

dandlezzz:
Upload files to ipfs via nft.storage's free service.

### DIFF
--- a/src/modules/ipfs/nft.storage/index.ts
+++ b/src/modules/ipfs/nft.storage/index.ts
@@ -6,7 +6,7 @@ export default async function uploadFile(file: File): Promise<PinFileResponse> {
   try {
     const response = await fetch("https://api.nft.storage/upload", {
       //@ts-ignore
-      body: fs.createReadStream(path),
+      body: fs.createReadStream(file.path),
       method: "POST",
       headers: {
         "Authorization": `Bearer ${process.env.NFT_STORAGE_API_KEY || ''}`,


src/modules/ipfs/client.ts
```diff
@@ -0,0 +1,36 @@
+import { Files } from 'formidable';
+import { PinFileResponse } from './types';
+import uploadFile from './nft.storage';
+import { fromDwebLink } from '.';
+
+export default async function UploadFiles(formFiles: Files ) {
```
kespinola, line 6: Since this isn't a class pattern is to camel case `uploadFiles`.

src/modules/ipfs/client.ts
```diff
@@ -0,0 +1,36 @@
...
+  const files = Object.values(formFiles)
+    .map(maybeFiles =>  maybeFiles instanceof Array ? maybeFiles[0] : maybeFiles)
```
kespinola, line 8:
```
const files = pipe(values, map(when(is(Array), take(0))))(formFiles)
```

ramda has some functions can help with readability.

src/modules/ipfs/client.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const uploadPromises: Promise<any>[] = []
+  files.forEach((file) => {
+    uploadPromises.push(uploadFile(file.path));
+  })
```
kespinola, line 13: I recommend doing this with a map.
```
const uploadPromises = map(pipe(pluck('path'), uploadFile))(files)
```

src/modules/ipfs/client.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const results = await Promise.allSettled(uploadPromises)
```
kespinola, line 15: allSettled first time seeing that one. 

```
The Promise.allSettled() method returns a promise that resolves after all of the given promises have either fulfilled or rejected, with an array of objects that each describes the outcome of each promise.
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

src/modules/ipfs/index.ts
```diff
@@ -0,0 +1,8 @@
+
+export const fromCloudFlare = (cid: string) => (
+  `https://cloudflare-ipfs.com/ipfs/${cid}`
+)
+
+export const fromDwebLink = (cid: string) => (
+  `https://${cid}.ipfs.dweb.link`
+)
```
kespinola, line 8:
```suggestion
export const fromCloudFlare = (cid: string):string => (
  `https://cloudflare-ipfs.com/ipfs/${cid}`
)

export const fromDwebLink = (cid: string):string => (
  `https://${cid}.ipfs.dweb.link`
)
```

The return type may be inferred but if not.

src/modules/ipfs/client.ts
```diff
@@ -0,0 +1,36 @@
...
+    const fileResponse: PinFileResponse = {
+      error: undefined,
+      uri: '',
+      name: files[index].name || '',
+      type: files[index].type || ''
+    }
+    if (result.status === "rejected"){
+      fileResponse.error = result.reason
+    } else {
+      const json = await result.value.json()
```
kespinola, line 28: Have some ideas on cleaning this up but involves resolving the json in the `uploadFile`. It should both make the request and parse into json.

Then this can be a transform without promises.

src/modules/ipfs/nft.storage/index.ts
```diff
@@ -0,0 +1,12 @@
+export default function uploadFile(path: string) {
```
kespinola, line 1:
```suggestion
export default function uploadFile(path: string):Promise<PinFileResponse> {
```

src/modules/ipfs/nft.storage/index.ts
```diff
@@ -0,0 +1,12 @@
...
+  return fetch("https://api.nft.storage/upload", {
+    //@ts-ignore
+    body: fs.createReadStream(path),
+    method: "POST",
+    headers: {
+      "Authorization": `Bearer ${process.env.NFT_STORAGE_API_KEY || ''}`,
+      "Content-Type": "application/x-www-form-urlencoded",
+    } 
+  })
```
kespinola, line 10: Yeah, this is the spot to do the await on the json. Have it return ready-to-go file responses.

kespinola (changes requested):
Some recommendations. Can chat if you aren't into them or have questions.

PR #74, opened 2021-10-08
Attempt to fix bug with loading Phantom.

ray-kast:


src/modules/wallet/provider.tsx
```diff
@@ -1,80 +1,92 @@
...
+        'load',
+        () => {
+          setSolana(window.solana);
+          setInitialization(false);
+        },
+        { capture: false }
+      );
     }
-  }, [])
+  }
```
ray-kast, line 52: The bulk of the changes happen here - the event handler for `load` does not seem to be called very often, so we optimistically look for an updated Phantom every time.

kespinola (approved):

PR #95, opened 2021-11-10
feat: nft info duplication and de-duplication of attribute types

kristianeboe:


src/modules/nfts/components/wizard/InfoScreen.tsx
```diff
@@ -113,6 +114,8 @@ export default function InfoScreen({
   dispatch,
   currentImage,
 }: Props) {
+  const [errors, setErrors] = useState<string[]>([]);
```
kespinola, line 117: Looks like there is a mix of using the validation engine from antd and this. Why wasn't a form.item validation sufficent?

kristianeboe: Aha, found the correct documentation for custom validators now, will implement that.

kespinola: sweet. ty.

kristianeboe: @kespinola I switched to the ant D custom validator (works great!), but had to keep the local error state. Otherwise it threw an error on every new attribute added (including the first one), instead of only validating on clicking Next.

kristianeboe:
Should be ready for you now David :)

src/modules/nfts/components/wizard/RoyaltiesCreators.tsx
```diff
@@ -332,14 +337,26 @@ export default function RoyaltiesCreators({
...
       });
   };
+  const removeCreator = (creatorAddress: string) => {
+    const newShareSplit = 100 / (creators.length - 1) || 100;
+    setCreators(
+      creators
+        .filter((c) => c.address !== creatorAddress)
+        .map((c) => ({ ...c, share: newShareSplit }))
+    );
+  };
```
kespinola, line 359: Also curious on why creator state isn't being tracked through the ant form?

Does setFieldsValue not suffice?

davetorbeck (approved):

kespinola (approved):
Let's progress and I'll slate some time after we release to see what refactors we implement. Easier to do once its all laid out.

The main item I want to review is our management of the form state. Like to leverage form api as much as possible.

PR #103, opened 2021-11-19
fix: nft creator percentage validation

kristianeboe:


kristianeboe:
We can probably refactor ( DRY up) even more stuff here, but this makes sure no one can proceed with percentages that don't add up to a 100%, and also carries over royalty settings from one NFT to another

kespinola (approved):

PR #105, opened 2021-11-19
fix: restored image duplicate handeling

kristianeboe:
If a user tries to upload a file with the same name(and file extension) as a previously uploaded file (upload meaning putting it into the browser with a file input tag, not IPFS) the system should allow it, but tack on a _1 (2,3..) at the end of the file name. The alternative being to just skip the "new" file.
Sooo: User tries to upload "image.png" two times, the system gets "image.png" and "image_1.png", and the minting process continues.

### DIFF
--- a/pages/nfts/new.tsx
+++ b/pages/nfts/new.tsx
@@ -99,6 +99,18 @@ const initialState: State = {
   nftValues: [],
 };
 
+function getFinalFile(file: File, numberOfDuplicates: number) {
+  // if there is a duplicate of file, add it as file_1 etc.
+  return numberOfDuplicates > 0
+    ? (() => {
+        const fileNameParts = file.name.split('.');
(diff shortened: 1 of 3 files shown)


pages/nfts/new.tsx
```diff
@@ -109,9 +109,26 @@ function reducer(state: State, action: MintAction) {
...
+
       return state.images.length < 10
-        ? { ...state, images: [...state.images, action.payload as File] }
-        : state; // since the dispactch is async(?) it did not reliably limit number of files from the Upload component.
+        ? {
+            ...state,
+            images: [
+              ...state.images,
+              nrOfDuplicates > 0
+                ? (() => {
```
davetorbeck, line 121: Can we extract out and name this function?

davetorbeck (approved):

PR #124, opened 2021-12-01
fix: Route to listing after successful NFT mint

davetorbeck:


kristianeboe (approved):
Nice fix! Could consider simply always redirecting to .com to "simplify" the code, up to you :)

PR #129, opened 2021-12-11
HOLA-1 add IPFS cdn

dandlezzz:
Leverage the brand new ipfs.cache.holaplex.com


kespinola (approved):
I pulled this logic and put in storefronts. It works we should ship.

PR #130, opened 2021-12-13
Discovery analytics

kristianeboe:
Hey all, ready for review!

- The filters are a little banged up right now, sorry, but I think you see where I'm going with it.
- Still some !important lying around in they styles
- There is some analytics stuff here too. Review it, or skip it, up to you!

Overall though, I'm pretty happy! Looking forward to see what you have to say :)


src/common/components/elements/CurrentListings.tsx
```diff
@@ -0,0 +1,290 @@
...
+}
+
+export function CurrentListings() {
+  const router = useRouter();
+  const [loading, setLoading] = useState(false);
+
+  //Example query ?search=hello&filters=active_auctions,&sort_by=ending_soonest
+  const queryFilters: string[] =
+    (router.query['filters[]'] as string[]) || (router.query.filters as string)?.split(',') || [];
+  const querySortBy = (router.query.sort as string) || 'recent_listings';
```
ray-kast, line 171: is `recent_listings` the right value here?

src/common/components/elements/ListingPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
+  ended: boolean;
+  last_bid?: number | null; // lamport price of last bid (thought it was a timestamp at first)
+  price_floor?: number | null;
+  total_uncancelled_bids?: number | null;
+  instant_sale_price?: number | null; // is often 10000000000
+  subdomain: string;
+  storeTitle: string;
+  items: Item[]; // NFT metadata, including URI to fetch more data
+  nftMetadata?: NFTMetadata[]; // same length as items. Is set on mount
+}
```
ray-kast, line 58: This issue is my fault, but I want to get all these fields changed to camelCase names on the backend before launch.

src/common/components/elements/ListingPreview.tsx
```diff
@@ -0,0 +1,343 @@
...
...dv22uIEiLA4q7nvvCug8WTqzQveOH26fodo7g6uFe/a17W3+nFBAkRYENRdb1vkkz1CH9cPsVy/jrhr27PqMYvENYNlHAIesRiBYwRy0V+8iXP8+/fvX11Mr7L7ECueb/r48eMqm7FuI2BGWDEG8cm+7G3NEOfmdcTQw4h9/55lhm7DekRYKQPZF2ArbXTAyu4kDYB2YxUzwg0gi/41ztHnfQG26HbGel/crVrm7tNY+/1btkOEAZ2M05r4FB7r9GbAIdxaZYrHdOsgJ/wCEQY0J74TmOKnbxxT9n3FgGGWWsVdowHtjt9Nnvf7yQM2aZU/TIAIAxrw6dOnAWtZZcoEnBpNuTuObWMEiLAx1HY0ZQJEmHJ3HNvGCBBhY6jtaMoEiJB0Z29vL6ls58vxPcO8/zfrdo5qvKO+d3Fx8Wu8zf1dW4p/cPzLly/dtv9Ts/EbcvGAHhHyfBIhZ6NSiIBTo0LNNtScABFyNiqFCBChULMNNSdAhJyNSiECRCjUbEPNCRAhZ6NSiAARCjXbUHMCRMjZqBQiQIRCzTbUnAARcjYqhQgQoVCzDTUnQIScjUohAkQo1GxDzQkQIWejUogAEQo121BzAkTI2agUIkCEQs021JwAEXI2KoUIEKFQsw01J0CEnI1KIQJEKNRsQ80JECFno1KIABEKNdtQcwJEyNmoFCJAhELNNtScABFyNiqFCBChULMNNSdAhJyNSiECRCjUbEPNCRAhZ6NSiAARCjXbUHMCRMjZqBQiQIRCzTbUnAARcjYqhQgQoVCzDTUnQIScjUohAkQo1GxDzQkQIWejUogAEQo121BzAkTI2agUIkCEQs021JwAEXI2KoUIEKFQsw01J0CEnI1KIQJEKNRsQ80JECFno1KIABEKNdtQcwJEyNmoFCJAhELNNtScABFyNiqFCBChULMNNSdAhJyNSiECRCjUbEPNCRAhZ6NSiAARCjXbUHMCRMjZqBQiQIRCzTbUnAARcjYqhQgQoVCzDTUnQIScjUohAkQo1GxDzQkQIWejUogAEQo121BzAkTI2agUIkCEQs021JwAEXI2KoUIEKFQsw01J0CEnI1KIQJEKNRsQ80JECFno1KIABEKNdtQcwJEyNmoFCJAhELNNtScABFyNiqFCBChULMNNSdAhJyNSiEC/wGgKKC4YMA4TAAAAABJRU5ErkJggg=="
```
ray-kast, line 290: Can this be included in the source tree as its own file instead?

src/common/constants/demoContent.ts
```diff
@@ -0,0 +1,131 @@
+// import listingsAndStorefronts from 'fixtures/demo-listings-and-storefronts.json';
```
ray-kast, line 1: is this file in use still?

src/modules/utils/callMetaplexIndexerRPC.ts
```diff
@@ -0,0 +1,58 @@
...
+    (listing?.price_floor || 0) * 0.000000001 > 2000 ||
+    storeBlacklist.includes(listing.subdomain)
+  ) {
+    return false;
+  }
+
+  return true;
+}
+
+export async function callMetaplexIndexerRPC(
```
ray-kast, line 23: This is most likely a ticket for later, but we should probably look at getting a library to do JSONRPC request and response handling.

ray-kast (changes requested):
I see that the Google Analytics IDs have been set to various amounts of `asdf`, was this intentional?

src/common/constants/demoContent.ts
kristianeboe: Removed :)

src/common/components/elements/CurrentListings.tsx
kristianeboe: Good catch, meant to be `recently_listed`

src/common/components/elements/ListingPreview.tsx
kristianeboe: It's like the last line of defence. In my testing so far I don't think I've ever seen it come up. We could consider removing it completely as well. @damianr Do you have any thoughts on a fallback image? We are using the standard from Antd now

image

pages/index.tsx
```diff
@@ -95,26 +117,26 @@ export default function Home({ featuredStorefronts }: HomeProps) {
...
-              grid={{ xs: 1, sm: 2, md: 3, lg: 3, xl: 3, xxl: 3, gutter: 16 }}
-              dataSource={drop<StorefrontFeature>(1, featuredStorefronts)}
-              renderItem={(feature) => (
-                <List.Item key={feature.storefront.subdomain}>
-                  <StorePreview {...feature} />
-                </List.Item>
-              )}
-            />
+          <Col xs={0} md={8}>
+            <FeaturedListingCarousel featuredListings={featuredListings} />
```
kespinola, line 121: For future work (and open for discussion) I don't like pulling things into seperate components until they are used in multiple spots. My personal ideal is just having the page component built out with component library.

src/common/components/elements/FeaturedListingsCarousel.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <span
+          style={{
+            color: 'white',
+            opacity: 0.6,
+            fontSize: 14,
+            fontWeight: 400,
+          }}
+        >
+          Featured listings
+        </span>
```
kespinola, line 25: Look to use text component with props like strong to build the text.

src/common/components/elements/FeaturedListingsCarousel.tsx
```diff
@@ -0,0 +1,57 @@
...
+            opacity: 1 !important;
+            padding-top: 5px !important;
+            padding-bottom: 5px !important;
+            border-radius: 100%;
+          }
+
+          .slick-active > button {
+            background: white !important;
+          }
+        `}
```
kespinola, line 53: Move this to a less sheet and import into the component.

kespinola: Try and wrap the carousel in an additional class and use that to override the specificity of the default styles without !important.

src/common/components/elements/ListingPreview.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+export function ListingPreview(listing: Listing) {
+  const storeHref = `https://${listing?.subdomain}.holaplex.com/#/auction/${listing.address}`;
+
+  const [showArtPreview, setShowArtPreview] = useState(false);
+  const [nft, setNFT] = useState<NFTMetadata | null>(null);
+
+  const nftMetadata = listing?.items?.[0]; // other items are usually tiered auctions or participation nfts
+  useEffect(() => {
+    async function fetchNFTDataFromIPFS() {
```
kespinola, line 255: We should work to get this as a method on the metadata class of the metaplex js sdk.

src/common/components/elements/ListingPreview.tsx
kespinola: I think you can put into another file and import the string.

kespinola: And yeah can just remove we're always going to have an image.

src/common/components/elements/CurrentListings.tsx
```diff
@@ -0,0 +1,323 @@
...
+        <DiscoveryFiltersAndSortBy
+          sortBy={state.sortBy}
+          filters={state.filters}
+          allFilterOptions={filterOptions}
+          allSortByOptions={sortingOptions}
+          dispatch={dispatch}
+        />
+      </Row>
+
+      <List
```
kespinola, line 284: What I'd like to work toward is having the components in the home page and wrapping the filter and infinite scroll into a hook.

src/common/components/elements/CurrentListings.tsx
```diff
@@ -0,0 +1,323 @@
...
+      payload: SortByAction;
+    };
+
+const initialState = (options?: {
+  filters?: FilterAction[];
+  sortBy?: SortByAction;
+}): DiscoveryToolState => {
+  const listingShells = Array(8)
+    .fill(null)
+    .map((_, i) => generateListingShell(i.toString()));
```
kespinola, line 113: What is the listing shell? Is it something we can drop seems to be putting together object.

kespinola (commented):
Nothing to worry about now but I'd like to do some refactor sessions post launch. We can adjust once filters and sorts are powered by the API.

Can you drop the fixture now?

kespinola (commented):
Screen Shot 2021-12-14 at 8 22 51 AM


Cases of the fallback showing up but the listing has artwork.

kespinola:
When scrolling the page it often blacks out.

kespinola:
Screen Shot 2021-12-14 at 8 30 22 AM


Before release lets wrapp the header in the ant pin component. Discussed this with @damianr .

kespinola:
When first loading the page the storefront and listing tiles are very small. then snap into place.

pages/index.tsx
kristianeboe: Yeah in general I agree. I created the component initially because I thought it would get more messy than it did 😅 Still for this one I recommend we keep the component as this section will probably be refactored in the future.

src/common/components/elements/CurrentListings.tsx
kristianeboe: I'll look to remove it. Used it initially to crate shadow listings, but might not be needed anymore

src/common/components/elements/FeaturedListingsCarousel.tsx
kristianeboe: I'll add it to the list! Might wait until after launch though 😅

src/modules/utils/callMetaplexIndexerRPC.ts
ray-kast: Yeah, I do recall discussing it, I was just swamped at the time.  Down to help find a library once I'm done with the backend.

src/common/components/elements/ListingPreview.tsx
kristianeboe: Done!

src/common/components/elements/CurrentListings.tsx
```diff
@@ -141,14 +140,15 @@ function reducer(state: DiscoveryToolState, action: DiscoveryToolAction): Discov
       };
     case 'LOAD_MORE_LISTINGS':
       const newCursor = state.cursor + pageSize;
+      const listings = new Set([
+        ...state.listingsOnDisplay,
+        ...state.filteredAndSortedListings.slice(state.cursor, newCursor),
+      ])
```
kristianeboe: @dandlezzz I'm not sure this can fix the root cause. new Set() only dedupes if the array is a list of primitives, not objects.

I created a codesandbox to verify: https://codesandbox.io/s/lingering-thunder-pwqmh?file=/src/App.tsx

I'll do some more tests on my end

kristianeboe: Turns out in this case it actually will work :O https://stackoverflow.com/questions/39997067/es6-unique-array-of-objects-with-set 

Because the object references are the same.

However, I think this has something to do with the cursor, so I'll look into that

kristianeboe: Solved it! Turns out I was resetting the cursor to 0 instead of page size on sort and filter. Sill!

src/common/components/elements/FeaturedListingsCarousel.tsx
kristianeboe: Solved

PR #132, opened 2021-12-15
update banner tooltip & preview aspect ratio

sjc5:
updates banner tooltip and also the aspect ratio of the banner in the preview card to match how it'll actually look on the storefront

related to https://github.com/holaplex/metaplex/pull/158

aspect ratio has great browser coverage now: https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio


sjc5:
closing temporarily but will revise and re-open once metaplex storefront revisions ship

PR #136, opened 2021-12-16
Builder analytics

kristianeboe:
- WIP
- Moving analytics logic over to a provider like in our metaplex
- currently experiencing a bug where gtag is not loaded during useEffect 🤔


kristianeboe:
If you want @kespinola I can take lead on merging this into dev on Monday. Some really cool concepts in here :)

PR #141, opened 2021-12-18
allow longer names & tweak form validation

sjc5:
lines up with recent changes to metaplex repo (https://github.com/holaplex/metaplex/commit/39bdc1c665982b567a42ff72975bf3cf02ebd164)

- allows longer NFT titles by relying on buffer / bytelength test instead of char limit
- adds 500 char max for description
- makes titles optional


kespinola (approved):
Nice. A bit more open for the change on this version of the minter. Way to close it all the way out.

PR #146, opened 2021-12-23
Kebo/builder analytics refactor

kristianeboe:


pages/_document.js
```diff
@@ -47,9 +48,9 @@ export default class MyDocument extends Document {
                       dataLayer.push(arguments);
                     }
                     gtag('js', new Date());
-                    gtag('config', '${this.props.googleAnalyticsId}');
-                    gtag('config', 'G-HLNC4C2YKN')
-                `,
+                    gtag('config', '${OLD_GOOGLE_ANALYTICS_ID}');
+                    `,
+                  // gtag('config',  '${GA4_ID}')
```
kristianeboe, line 53: we are initializing the new gtag in the AnalyticsProvider

src/modules/ganalytics/AnalyticsProvider.tsx
```diff
@@ -1,107 +1,177 @@
...
+      });
+    }
+  }
 
-  function setUserProperties(attributes: AnalyticsUserProperties) {
-    gtag('set', 'user_properties', {
-      ...attributes,
-    });
+  function identify() {
+    console.log(pubkey ? 'User identified' : 'No user identified');
```
kespinola, line 119: remove. We should add no console log to our linting.

src/modules/ganalytics/AnalyticsProvider.tsx
```diff
@@ -111,68 +181,52 @@ export function AnalyticsProvider(props: { children: React.ReactNode }) {
             value,
           }),
       ...otherAttributes,
-    });
+    };
+
+    console.log('track', action, attrs);
```
kespinola, line 186: remove.

src/modules/ganalytics/AnalyticsProvider.tsx
```diff
@@ -1,107 +1,177 @@
...
+          if (pubkey) {
+            event.setUser(pubkey);
+          }
+          if (false) {
+            // for later inspiration
+            event.addMetadata('company', {
+              name: 'Acme Co.',
+              country: 'uk',
+            });
+          }
```
kespinola, line 112:
```suggestion
```

If its not used then remove.

pages/index.tsx
```diff
@@ -412,20 +456,27 @@ export default function Home({ featuredStorefronts }: HomeProps) {
...
+              {take(show, displayedListings).map((listing: Listing, i) => (
                 <Col xs={24} sm={12} md={8} lg={8} xl={6} xxl={6} key={listing?.listingAddress}>
-                  <ListingPreview {...listing} />
+                  <ListingPreview
+                    listing={listing}
+                    meta={{
+                      index: i,
+                      list: 'current-listings',
+                      sortBy: sortBy,
+                      filterBy: filterBy,
```
kespinola, line 477: Instead of passing sortBy and filterBy to each preview look to decorate a track function in this view and pass that to preview.

```
const listingTrack = (args...) => track({...args, sortBy, filterBy })
```

src/modules/ganalytics/AnalyticsProvider.tsx
```diff
@@ -187,3 +241,63 @@ export function useAnalytics() {
...
   return context;
 }
+
+const listNames = {
+  featuredListings: 'Featured listings',
+  currentListings: 'Current listings',
+};
+
+// static function to augment track event
+export function addListingToTrackCall(listing: Listing) {
```
kespinola, line 251: My opinion is to remove all the info from the event that we can pulled from the database. If we want to do an analysis of listings we are going to go to the db not google analytics. 

I do see potential for faceting events of how many click throughs for buy now vs auctions where current bid is 0.5 so mixed so staying open to possibilities.

src/modules/ganalytics/AnalyticsProvider.tsx
```diff
@@ -1,107 +1,177 @@
...
+
+    return () => {
+      router.events.off('routeChangeComplete', pageview);
+    };
+  }, [router.events]);
+
+  function track(action: AnalyticsAction, attributes: TrackingAttributes) {
+    const { category, label, value, sol_value, ...otherAttributes } = attributes;
+
+    const attrs = {
```
kespinola, line 170: Mentioned this before but the track function is becoming bloated. It should only have the attributes that are going to be on every event. And then in the page handler decorate track with specifics for that page and use the decorated track event when it comes time to fire an event.

src/modules/ganalytics/AnalyticsProvider.tsx
kespinola: I'd like to see the event prep happening in the view leveraging track. Don't want to overload the analytics provider knowledge of every entity making use of it.

src/modules/ganalytics/AnalyticsProvider.tsx
```diff
@@ -1,107 +1,177 @@
...
+
 import { Listing } from '@/modules/indexer';
+import { useRouter } from 'next/router';
+import {
+  getFormatedListingPrice,
+  lamportToSolIsh,
+} from '@/common/components/elements/ListingPreview';
 
-export const GOOGLE_ANALYTICS_ID = process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS_ID || 'G-HLNC4C2YKN';
+export const OLD_GOOGLE_ANALYTICS_ID = process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS_ID;
```
kespinola, line 15: I'm for making a clean switch. Just say on the release date all data will now be going to update google tracking.

src/common/components/elements/ListingPreview.tsx
```diff
@@ -275,20 +301,28 @@ export function ListingPreview(listing: Listing) {
...
   if (loading || !listing.subdomain) {
     return <SkeletonListing />;
   }
 
   return (
-    <div ref={cardRef}>
+    <div
+      ref={cardRef}
+      onClick={() => {
+        track('Select listing', {
```
kespinola, line 315: The referrer not enough? Click throughs out of the site into a storefront are already tracked by the storefronts tracking as a referrer.

kespinola (changes requested):
I'm for deprecating the old events method vs accommodating both.

src/modules/ganalytics/AnalyticsProvider.tsx
kristianeboe: Got it, I'll clean up all the console logs :)

pages/index.tsx
kristianeboe: Yeah, that's a good idea. Will do!

Do you mind if we keep passing down listing as a named prop though? I think it's cleaner than the {... listing} syntax :shrug:

src/modules/ganalytics/AnalyticsProvider.tsx
kristianeboe: I'd like to keep the additional listing event properties for now if you don't mind too much. I agree serious (money/sol) analytics should happen at database level, but I really want to  experiment with those extra properties in GA.

Also agree to move the helper function out of the analytics provider for now. It could make sense to add it back in if we started tracking listings from multiple views in the future though.

src/modules/ganalytics/AnalyticsProvider.tsx
kristianeboe: There is no penalty for running it in parallel. I guess I'm a bit afraid of something not working in prod and us losing data from the initial/old implementation.

I can make a ticket and a PR for moving "storefront builder" analytics to the new provider after this one is merged. Would that work for you @kyle? :)

src/common/components/elements/ListingPreview.tsx
kristianeboe: I think using only the referrer leaves potentially valuable data on the table, and I think it will also be easier to build reports on this event, put it into funnels etc.

I really want to keep it

kespinola (approved):
Thanks for cleaning up the console logs, unused code, and responding back to my comments.

PR #148, opened 2021-12-27
Bulk minter small changes

davetorbeck:


pages/nfts/new.tsx
```diff
@@ -224,61 +224,59 @@ export default function BulkUploadWizard() {
...
+    const category: NFTCategory = detectCategoryByFileExt(filePin.name);
+    const isMultiMedia = category !== 'image';
 
-        let coverImageFile: UploadedFilePin | undefined;
-        if (v.coverImageFile) {
-          coverImageFile = await uploadCoverImage(v.coverImageFile);
-        }
+    let coverImageFile: UploadedFilePin | undefined;
+    if (value.coverImageFile) {
+      coverImageFile = await uploadCoverImage(value.coverImageFile);
```
davetorbeck, line 243: @kespinola When you say this should be moved out of this transform, do you have an idea where?  As in we should store the filePins for coverImages earlier then retrieve them based on index?

src/common/components/elements/ModelViewer.tsx
```diff
@@ -3,7 +3,7 @@ import '@google/model-viewer';
 
 const ModelViewer = (props: any) => (
   // @ts-ignore
-  <model-viewer {...props} style={{ width: 800, height: 400 }}></model-viewer>
+  <model-viewer {...props} style={{ width: '60vw', height: '30vw' }}></model-viewer>
```
kespinola, line 6: I'd like to see this be width: 100% and then define some width based media queries to adjust the height.

pages/nfts/new.tsx
davetorbeck: So my thought was when they select a cover image on the info page they may hit the back button and make a change to that image, so we only upload the coverImage to IPFS when we are about to send off the metadata and avoid putting things on IPFS unless absolutely needed.

src/common/components/elements/ModelViewer.tsx
davetorbeck: We want the preview modal to take the full width of the viewport?

davetorbeck: So we want the width to be 100% and then set the width directly on the modal that contains it?  This is a bit trickier because of the cascade and needing to target multiple antd classnames in the modal.  I can move out the explicit settings up to to where the component is called and keep this component non-opinionated.

davetorbeck:
Merging since needed for pulling out the minting flow to modalize.  Can handle open Qs in a later PR.

PR #149, opened 2021-12-28
feat: removed useWindowDimensions from app header

kristianeboe:
- moved displaying of Holoaplex and Edit store in AppHeader away from useWindowDimensions hook and into Styled component (CSS)
- updated show prop to $show to make it a "transient" prop that only goes to the styled component and not the DOM. https://styled-components.com/docs/api#transient-props
- Tried to delete the whole useWindowDimensions file, but ended up with a weird module not found error, so left it in for now.

### DIFF
--- a/src/common/components/elements/ListingPreview.tsx
+++ b/src/common/components/elements/ListingPreview.tsx
@@ -49,8 +49,8 @@ const Square = styled(Row)`
   }
 `;
 
-const NFTPreview = styled(Image)<{ show: boolean }>`
-  display: ${({ show }) => (show ? 'block' : 'none')};
+const NFTPreview = styled(Image)<{ $show: boolean }>`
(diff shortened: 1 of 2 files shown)


kristianeboe:
image
image

Look, no more errors in the console :)

src/common/components/elements/ListingPreview.tsx
```diff
@@ -288,7 +288,7 @@ export function ListingPreview(listing: Listing) {
         <ListingPreviewContainer>
           <Square>
             <NFTPreview
-              show={inView}
+              $show={inView}
```
kespinola, line 291: Not sure on the $ notation where did you pick that up from?

kespinola (approved):

sjc5 (approved):
Looks good and nice touch hiding "Edit store" on mobile view.  Maybe we should capitalize the word "store" in that button while we're at it?

PR #150, opened 2021-12-28
Metadata signing via Rabbit MQ

dandlezzz:
This PR brings our metadata signing to async processing coordinated via Rabbit MQ.

Via Rabbit MQ we can defer signing NFT metadata and take it outside of the web request flow. This will help accommodate what can occasionally be a pretty long wait for the Solana transaction to confirm.

We have one queue for signing, if the signing fails it goes into a "dead-letter" queue. From the rabbit mq interface we can requeue jobs, multiple at once, but manually. Eventually, we can add automatic retries. 

 How to run the consumer `yarn run consumers:sign-metadata`.
This will spin up a consumer in the background. All signing taks will go to the queue from here on out.


kristianeboe (approved):
Clean code!

tasks/consumers/sign-metadata.ts
```diff
@@ -0,0 +1,152 @@
...
+      } catch {
+        console.error('Invalid public keys');
+        channel.reject(msg, false)
+        return;
+      }
+
+      try {
+        const tx = new Transaction();
+
+        signMetadata(metadata, keypair.publicKey, tx, metaProgramId);
```
kespinola, line 125: The sdk has this action. Can we switch to that at this time?

src/modules/metadata-signing.ts
```diff
@@ -0,0 +1,2 @@
+
+export const signingQueue = "metadata-signging"
```
kespinola, line 2: Move this to an environment variable?

tasks/consumers/sign-metadata.ts
dandlezzz: I'd like to address that in a separate PR.

src/modules/metadata-signing.ts
dandlezzz: It's better if these don't change without a code update. The assertQueue method creates a queue if one doesn't exist by that name. Don't really want them changed too often.

PR #156, opened 2021-12-30
Use SVG for SOL symbol instead of unicode character

### LINKED ISSUE #151: Use SVG for SOL icon instead of Unicode char (◎)
**Problem:**

- Current SOL icon vertically aligns differently across browsers in an annoying way.

**Goal:**

- Use an SVG icon instead.

**Questions:**

- Does it make more sense to use the S style Solana logo or the double-circle style logo?  What seems to be more prevalent across community?  @damianr @bellebellesol

sjc5:
Uses SVG for SOL symbol instead of unicode character

Fixes #151 

**Previews:**

Screen Shot 2021-12-30 at 11 44 10 AM

Screen Shot 2021-12-30 at 11 44 02 AM


src/modules/nfts/components/wizard/PriceSummary.tsx
```diff
@@ -104,9 +105,16 @@ export default function PriceSummary({
...
-                </Paragraph>
+                <div
+                  style={{
+                    fontSize: 14,
+                    display: 'flex',
+                    alignItems: 'center',
+                    marginTop: '-0.9rem',
+                  }}
+                >
+                  <SolSymbol size={14} /> <span>{SOL_COST_PER_NFT}</span>
```
kespinola, line 116: I recommend wrapping the sol symbol in a styled component. Apply these extra styles to it.

src/modules/nfts/components/wizard/PriceSummary.tsx
```diff
@@ -115,9 +123,16 @@ export default function PriceSummary({
...
-                  <span className="sol-icon">◎</span> {totalSolCost}
-                </Paragraph>
+                <div
+                  style={{
+                    fontSize: 18,
+                    display: 'flex',
+                    alignItems: 'center',
+                    marginTop: '-0.25rem',
+                  }}
+                >
```
kespinola, line 133: Same on wrapping or add a class to the the symbol component and target with a styled component already wrapping the symbol.

kespinola (commented):
Recommendations on handling the styles with out style prop.

src/modules/nfts/components/wizard/PriceSummary.tsx
```diff
@@ -1,16 +1,17 @@
...
 import Button from '@/common/components/elements/Button';
+import SolSymbol from '@/common/components/elements/SolSymbol';
 import Paragraph from 'antd/lib/typography/Paragraph';
 import { Coingecko, Currency } from '@metaplex/js';
 import NavContainer from '@/modules/nfts/components/wizard/NavContainer';
 import { WalletContext } from '@/modules/wallet';
 import { Connection, LAMPORTS_PER_SOL, PublicKey } from '@solana/web3.js';
 import { NFTPreviewGrid } from '@/common/components/elements/NFTPreviewGrid';
 import { FilePreview } from 'pages/nfts/new';
+import { Flex } from 'antd-mobile';
```
kristianeboe, line 14: @sjc5 Are we using antd-mobile in many other places, or will we start doing so? :)

sjc5: hmm, I think that was an auto-import error, as I don't remember adding it.  let me double check.

src/modules/nfts/components/wizard/PriceSummary.tsx
sjc5: Yeah good idea, will do that in both cases

src/common/components/elements/Price.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <path
+          d="M360.905 273.786c.711 0 1.319.252 1.825.756a2.48 2.48 0 0 1 .759 1.822 2.48 2.48 0 0 1-.759 1.823 2.493 2.493 0 0 1-1.825.756 2.488 2.488 0 0 1-1.82-.756 2.48 2.48 0 0 1-.758-1.823c0-.714.252-1.323.755-1.825a2.49 2.49 0 0 1 1.823-.753Zm0 4.717c.586 0 1.089-.21 1.508-.63.42-.42.63-.923.63-1.509a2.06 2.06 0 0 0-.627-1.511 2.058 2.058 0 0 0-1.511-.627c-.582 0-1.083.21-1.503.63-.42.42-.63.923-.63 1.508 0 .586.21 1.089.63 1.509.42.42.921.63 1.503.63Zm0-3.422c.355 0 .659.126.911.378s.378.554.378.905c0 .356-.126.659-.378.909s-.556.375-.911.375c-.352 0-.654-.126-.906-.378a1.236 1.236 0 0 1-.378-.906c0-.351.126-.653.378-.905s.554-.378.906-.378Zm0 2.121a.818.818 0 0 0 .597-.243.804.804 0 0 0 .246-.595.803.803 0 0 0-.246-.594.818.818 0 0 0-.597-.243.812.812 0 0 0-.592.243.803.803 0 0 0-.246.594c0 .235.082.433.246.595a.812.812 0 0 0 .592.243Z"
+          style={{
+            fillRule: 'nonzero',
+            fill: 'currentcolor',
+          }}
+          transform="translate(-10120.482 -7732.751) scale(28.2437)"
+        />
+      </svg>
+      <span style={{ fontSize: props.size + 'px' }}>{props.price}</span>
```
kespinola, line 39: You can use props within styled components. use that to set the font sizes.

here is an example

https://github.com/holaplex/holaplex/blob/dev/pages/index.tsx#L152

kespinola (changes requested):

kespinola (approved):

sjc5:
including via #161 instead

PR #161, opened 2022-01-07
Quick disco style updates

### LINKED ISSUE #151: Use SVG for SOL icon instead of Unicode char (◎)

sjc5:
- Changes holaplex.com font to Inter (from Nunito)
- Makes black background into dark grey and white font color ever-so-slightly softer to be easier on the eyes
- Adjust font weights as needed
- Slightly increase left / right margin at small laptop / tablet break points (no changes on mobile and large screens)
- Uses SVG for sol symbol instead of a unicode character
- Sets Inter as the default font for storefronts in the store setup/edit flow
- Adds font smoothing
- various other minor adjustments

All per discussions with @bellebellesol

Resolves #151

### DIFF
--- a/ant-theme.less
+++ b/ant-theme.less
@@ -3,7 +3,7 @@
 
 @base-background: #070707;
 @font-size-base: 18px;
-@text-color: white;
+@text-color: #eaeaea;
 @border-radius-base: 5px;
 @primary-color: #f03e92;
 @font-family: 'Inter', -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Oxygen, Ubuntu,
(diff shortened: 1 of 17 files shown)


kristianeboe (approved):
Looks good! Wish we had a way to easily  standardise style variables across styled components too... oh well, another time

kespinola (approved):
looking good by me. This is all @damianr .

PR #168, opened 2022-01-14
Save CrossMint ID to Arweave

sjc5:
Creates a CrossMint user id by calling uuid when a user without an existing crossmint ID creates a new store or edits their existing store, and persists the CrossMint ID to arweave.  This is required to enable CrossMint credit and debit card payments on storefronts.


pages/storefront/edit.tsx
```diff
@@ -77,6 +79,10 @@ export default function Edit() {
     },
     { name: ['meta', 'title'], value: storefront?.meta.title ?? '' },
     { name: ['meta', 'description'], value: storefront?.meta.description ?? '' },
+    {
+      name: ['integrations', 'crossmintClientId'],
+      value: storefront?.integrations?.crossmintClientId ?? uuidv4(),
```
kristianeboe, line 84: Does the process fail if it doesn't have a value? Is the randomized value helpful in other regards? Just curious :)

alfonso-paella: From crossmint perspective, we do need the client ID as it's what's used to enable the tool on our backend.

If it failed, we'd need a manual way to backfill it or else have the client create a new storefront, which feels veryy painful >.<

kespinola (approved):

sjc5:
thanks @kespinola!

PR #181, opened 2022-01-21
Kebo/secondary badge

kristianeboe:

PR #197, opened 2022-01-28
Activity Page

kevinrodriguez-io:


src/modules/utils/string.ts
```diff
@@ -0,0 +1,2 @@
+export const showFirstAndLastFour = (str: string, isLowerThanEight = str.length <= 8) =>
```
kristianeboe, line 1: Is this used for anything else than truncating pub keys? Maybe we should rename it to truncatePubkey or shortenPubkey?

src/common/hooks/useTwitterHandle.ts
```diff
@@ -0,0 +1,24 @@
+import { getHandleAndRegistryKey } from '@solana/spl-name-service';
```
kristianeboe, line 1: This looks nice! I might port this over to metaplex as well

src/common/styles/ButtonReset.ts
```diff
@@ -0,0 +1,11 @@
+import { css } from 'styled-components';
```
kristianeboe, line 1: We are pulling in tailwind in this branch https://github.com/holaplex/holaplex/pull/181

Are you using Styled components many places? Would you like to switch it over to TW now or just refactor it later?

src/common/hooks/useOutsideAlerter.ts
```diff
@@ -0,0 +1,15 @@
+import { RefObject, useEffect } from 'react';
```
kristianeboe, line 1: Is this one from here :) ? https://usehooks.com/useOnClickOutside/

kristianeboe (changes requested):
Looks good, just a few comments and questions

src/common/styles/ButtonReset.ts
kevinrodriguez-io: heheh... A bit late tho. Let's do it later because I made a ton of components using styled.

src/common/components/elements/ActivityContent.tsx
```diff
@@ -0,0 +1,452 @@
...
+          <LoadingActivitySkeletonBox />
+        </>
+      ) : hasItems ? (
+        <>
+          {items.map((bid, i) => (
+            <ActivityBox
+              key={i}
+              disableMarginTop={i === 0}
+              relatedImageUrl={
+                maybeImageCDN(bid.listing?.nfts?.[0]?.image) ??
```
kespinola, line 106: We'll get the api returning the cdn version of the image so enough.

src/common/hooks/useOutsideAlerter.ts
kevinrodriguez-io: I think I found it elsewhere

src/common/hooks/useLamportBalance.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const useLamportBalance = () => {
+  const { publicKey } = useWallet();
+  const { connection } = useConnection();
+  return useSWR(['lamport-balance', publicKey?.toBase58()], (_: string, pk: string) => {
+    if (pk && connection) {
+      return connection.getBalance(new PublicKey(pk));
+    }
+  });
+};
```
kespinola, line 13: nice.

kespinola (changes requested):
Only blocker is the adjustment of the route to `/profiles/[wallet]`. Only because once routes go out they spread so want to get the route naming spot on before release.

PR #214, opened 2022-02-10
Changes and add toast

kevinrodriguez-io:


sjc5 (approved):
Made some fixes / improvements, mainly removing the appHeader bottom margin JS logic and replacing with CSS, plus some other CSS edits (profile page L/R margins + converting inline styles to tailwind classes). Also fixed a conditional rendering bug and removed some dev/debug code.  Otherwise looking awesome.  Approved

Issue #200, opened 2022-02-03
Storefront JSON Update Script


kespinola:
## Goal
Holaplex engineers can patch storefront json of a storefront by executing a yarn command on a server with access to arweave private key.

## Changes
- Add a node script to hacks directory. [exp script](https://github.com/holaplex/holaplex/blob/dev/hacks/demo-theme.ts)
```
yarn run storefronts:update /path/to/storefront.json
```
- Script reads the json file provided, passes results to upsert arweave storefront sdk. 

https://github.com/holaplex/holaplex/blob/dev/src/modules/arweave/client.ts#L237

_The css string can be empty. No longer used. Ideally save the json contents to the data_

kespinola:
done.

PR #216, opened 2022-02-11
storefronts JSON update script

sjc5:
adds a script and example storefront json file so an engineer logged into a server with access to private arweave key can update a storefront.

solves #200


hacks/update-store-json.ts
```diff
@@ -0,0 +1,87 @@
...
+  const parseStorefront = ajvParse(schemas.parser(SCHEMAS.storefront));
+
+  const payloadRes = await resultThenAsync(parseNotarized<Storefront>(params), (params) =>
+    unpackNotarized(
+      params,
+      verifyNaclSelfContained((s) => new PublicKey(s.pubkey).toBuffer()),
+      { parse: parseStorefront, format: formatMessage }
+    )
+  );
+
```
kespinola, line 32: We don't need the notarize these since ran by us. This is for publically exposed uploads.

kespinola (changes requested):

PR #224, opened 2022-02-22
Migrate to Wallet Adaptor

kespinola:
## Changes
- Replace custom phantom only wallet connecting with wallet adapter.
- Move to pattern of looking up storefront and upserting wallet into our db as soon as user connects.
- 2-step process for seeing create/edit storefront button now. 
- Move nfts model to nfts/new. Use router.back to get back experience of rendering in base _app component.
- Revert WhiteRounded buttons @sjc5 can you update the existing button component? We have no need for the pink anymore so all buttons should be rounded. The ant theme primary does white and secondary is black.

Screen Shot 2022-02-21 at 4 15 08 PM
Screen Shot 2022-02-21 at 4 15 49 PM
Screen Shot 2022-02-21 at 4 16 18 PM
Screen Shot 2022-02-21 at 4 17 18 PM


pages/index.tsx
```diff
@@ -475,21 +482,32 @@ export default function Home({ featuredStorefronts, selectedDaoSubdomains }: Hom
...
-              {MARKETPLACE_ENABLED && (
-                <WhiteRoundedButton onClick={() => router.push('/marketplace/new')}>
-                  Create Your Marketplace
-                </WhiteRoundedButton>
-              )}
-            </Space>
+            {connected ? (
+              <Space direction="horizontal" size="large">
+                {searching ? (
+                  <Skeleton.Button></Skeleton.Button>
```
kespinola, line 488: When the final button dimensions are in place this should be the same size and shape as the actual button so screen doesn't move while loading. @sjc5

pages/marketplace/new.tsx
```diff
@@ -1,13 +1,12 @@
-import { programs } from '@metaplex/js';
+import { programs, Wallet } from '@metaplex/js';
 import DomainFormItem from '@/common/components/elements/DomainFormItem';
-import { WhiteRoundedButton } from '@/components/elements/Button';
+import Button from '@/components/elements/Button';
```
kespinola, line 3: I really like the loading props for the ant button. Instead of switching to vanilla button can you look to theme the existing ant button? You can strip out all the pink stuff. We are just going with the white for primary and black for scondary with all having the full rounded corners.

Do try to tailwind but thinking something like this is the move.

https://holaplex.slack.com/archives/C02DZRTN6TZ/p1645206614278599

pages/profiles/[wallet].tsx
```diff
@@ -11,15 +11,13 @@ import { PublicKey } from '@solana/web3.js';
...
-import { useWallet } from '@solana/wallet-adapter-react';
-import { useRouter } from 'next/router';
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
   // ...
   return {
     props: {
       // query params must be gotten serverside to be available on initial render
-      wallet: context?.params?.wallet ?? '',
+      wallet: context.query.wallet,
```
Comment (kespinola) [inline pages/profiles/[wallet].tsx:20]: Don't think I this should have been changed. Needs double check.

src/modules/ganalytics/AnalyticsProvider.tsx
```diff
@@ -77,8 +77,10 @@ export function AnalyticsProvider(props: { children: React.ReactNode }) {
   const router = useRouter();
   const [trackingInitialized, setTrackingInitialized] = useState(true);
   const [trackingAccepted, setTrackingAccepted] = useState(true);
-  const { wallet, publicKey } = useWallet();
-  const pubkey = publicKey?.toString() || '';
+  const { wallet } = useContext(WalletContext);
+  const pubkey = wallet?.pubkey || '';
+  // const pubkey = publicKey?.toBase58() || '';
+  // const endpointName = ENDPOINTS.find((e) => e.endpoint === endpoint)?.name;
```
kespinola, line 83: needs to be deleted.

kespinola (commented):
@sjc5 when you are next free love some help with the items I commented on. Thanks.

PR #258, opened 2022-03-04
Remove collection support until ready

davetorbeck:


dandlezzz (approved):
we'll bring it back soon!

PR #275, opened 2022-03-24
Nft page

davetorbeck:


pages/nfts/[address].tsx
```diff
@@ -1,8 +1,50 @@
...
+import Accordion from '../../src/common/components/elements/Accordion';
 // import Bugsnag from '@bugsnag/js';
 
+const Avatar = ({ address }: { address: string }) => {
+  const { data: twitterHandle } = useTwitterHandle(null, address);
+  console.log('twitterHandle', twitterHandle);
+  const [queryWalletProfile, { data }] = useWalletProfileLazyQuery();
+
+  useEffect(() => {
+    if (!twitterHandle) return;
```
Comment (kristianeboe) [inline pages/nfts/[address].tsx:22]: What is going on here? It should probably be the other way around, trying graphQL first, and then useTwitterHandle later if graphql did not find it

pages/nfts/[address].tsx
```diff
@@ -12,10 +54,9 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
 };
 
 export default function NftByAddress({ address }: { address: string }) {
-  const [queryNft, nftResponse] = useNftPageLazyQuery();
-
-  const nft = nftResponse?.data?.nft;
-  console.log('nft', nft);
+  const [queryNft, { data, loading, called }] = useNftPageLazyQuery();
```
Comment (kristianeboe) [inline pages/nfts/[address].tsx:57]: is it possible to move this to the server? I guess it won't be a  lazy query anymore. Hmmm

pages/nfts/[address].tsx
```diff
@@ -32,9 +73,112 @@ export default function NftByAddress({ address }: { address: string }) {
...
+                <h1 className="mb-4 text-2xl">{nft?.name}</h1>
+                <p className="text-lg">{nft?.description}</p>
+              </>
+            )}
+          </div>
+          {loading ? (
+            <div className="aspect-square w-full rounded-lg border-none bg-gray-800" />
+          ) : (
+            <img
+              src={nft?.image}
```
Comment (kristianeboe) [inline pages/nfts/[address].tsx:98]: add in the imgOpt function here, like in ListingPreview

kristianeboe (changes requested):
Great work so far! Left some comments and suggestions :) Also some console.logs, but it is a draft after all :)

pages/nfts/[address].tsx
```diff
@@ -12,10 +54,9 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
-  const [queryNft, nftResponse] = useNftPageLazyQuery();
-
-  const nft = nftResponse?.data?.nft;
-  console.log('nft', nft);
+  const [queryNft, { data, loading, called }] = useNftPageLazyQuery();
+  const nft = data?.nft;
+  // const isOwner = equals(data?.nft.owner.address, publicKey?.toBase58()) || null;
 
   useEffect(() => {
     if (!address) return;
```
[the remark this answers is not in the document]
Comment (davetorbeck) [reply pages/nfts/[address].tsx:113]: Not sure I follow?  Why does it need to be reactQuery?

pages/nfts/[address].tsx
Comment (davetorbeck) [reply pages/nfts/[address].tsx:108]: We could but sometimes I feel the query is slow and it might be better to insta-load the page and have it resolve since we have UI placeholders.

pages/nfts/[address].tsx
Comment (davetorbeck) [reply pages/nfts/[address].tsx:73]: Not sure how we get twitterHandles off of graphql? Is there one supported?  This is the way its done in profiles so I assume it was default

image

PR #277, opened 2022-03-25
moving assets subdomain to holaplex.tools

mpwsh:


mpwsh:
Please contact me before merging, we need to reconfigure the cdn in sync to this change to avoid downtime.

PR #293, opened 2022-04-04
Off site links open in new page

echohtp:
Small style suggestion - links to off domain sites should open in new tab.


kristianeboe:
Thanks @echohtp!

PR #308, opened 2022-04-05
Graphql engine + Moving counters to graphql and profile overview + Other goodies

kevinrodriguez-io:


codegen-apollo.yml
```diff
@@ -1,4 +1,7 @@
-schema: ${NEXT_PUBLIC_INDEXER_GRAPHQL_URL}
+schema:
+  - ${NEXT_PUBLIC_HASURA_GRAPHQL_URL}:
+      headers:
+        x-hasura-admin-secret: ${HASURA_GRAPHQL_ADMIN_SECRET}
```
kristianeboe, line 4: Is this public on the web client in the end?

package.json
```diff
@@ -122,6 +122,7 @@
     "autoprefixer": "^10.4.2",
     "babel-plugin-import": "^1.13.3",
     "chai": "^4.3.4",
+    "dotenv-flow": "^3.2.0",
```
kristianeboe, line 125: I thought regular dotenv already worked well with multiple files? I think it goes in the order of
.env
.env.development
.env.local

The difference maybe being that variables can't be overwritten
A solution could be to just # out the variables in .env that needs to be overwritten by .local

src/common/components/elements/FollowerBubble.tsx
```diff
@@ -3,60 +3,32 @@ import Link from 'next/link';
...
-  }, [fromHandle, queryWalletProfile]);
-
-  const isLoading = status === 'loading' || status === 'idle' || result.loading;
-
   return (
-    <Link href={`/profiles/${follower.account.from.toBase58()}`} passHref>
+    <Link href={`/profiles/${follower.pubKey}`} passHref>
       <a
         className={cx({
           'ml-[-8px] block': !isFirst,
```
kristianeboe, line 19: You can use "-space-x-2" on the parent div instead of the pr element styling. 

Here is a full example
image

 <div ```
className="flex -space-x-2 overflow-hidden">
        
        
        
        
      </div>
```

**Review by kespinola — CHANGES_REQUESTED** [review]
We talked at standup:
- Move connection querying to graph.holaplex.com
- Also with Hasura inconsistent response times I don't feel comfortable shipping it to our products at this time. For now should use as a prototype api for newly indexed programs. We need to audit the performance and make adjustments like vpc linking before we make it our api gateway (which Im excited about)

File: codegen-apollo.yml (hunk 1 above)
Comment (kevinrodriguez-io) [reply codegen-apollo.yml:4]: I don't understand, but right now queries would go through hasura, however we may end up changing this and have two different clients.

File: package.json (hunk 2 above)
Comment (kevinrodriguez-io) [reply package.json:125]: Yeah It kinda doesn't for the .env.local with the dotenv thing for the yaml configs

File: pages/_app.tsx [hunk 4]
```diff
@@ -8,7 +8,7 @@ import Head from 'next/head';
 import styled from 'styled-components';
 import { Layout } from 'antd';
 import { isNil } from 'ramda';
-import { WalletProvider } from '@/modules/wallet';
+import { WalletProviderDeprecated } from '@/modules/wallet';
```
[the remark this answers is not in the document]
kevinrodriguez-io: :D

kevinrodriguez-io:
@kespinola sure I'd be switching to the new endpoints. my only request would be a simple endpoint for the profile overview to have it ready via ssr
is that possible?
#import "./profileInfo.graphql"

query getProfileFollowerOverview($pubKey: String!) {
  to: graph_connections_aggregate(where: { from_account: { _eq: $pubKey } }) {
    aggregate {
      count
    }
  }
  from: graph_connections_aggregate(where: { to_account: { _eq: $pubKey } }) {
    aggregate {
      count
    }
  }
  to_brief: graph_connections(
    where: { to_account: { _eq: $pubKey } }
    limit: 4
    # To-do: Implement a closeness interaction graph or some sort of scoring system.
    order_by: { from_account_twitter_handle_name_services: { twitter_handle: asc } }
  ) {
    pubKey: from_account
    profileInfo: from_account_twitter_handle_name_services {
      ...ProfileInfo
    }
  }
}

fragment ProfileInfo on twitter_handle_name_services {
  twitter_handle
  wallet_address
  images: twitter_handle_profile_data {
    bannerImageUrl
    profileImageUrlHighres
    profileImageUrlLowres
  }
}

kevinrodriguez-io:
@kespinola Switched endpoints! Thanks !

pages/_app.tsx
```diff
@@ -138,7 +130,7 @@ function MyApp({ Component, pageProps }: AppProps) {
           <ConnectionProvider endpoint={endpoint}>
             <WalletProviderSolana wallets={wallets} autoConnect>
               <WalletModalProvider>
-                <WalletProvider>
+                <WalletProviderDeprecated>
```
kespinola, line 133: I don't want to further complicate our stack with 2 graphql clients. Lets just use our graph endpoint for now but will make the switch as hasura for the gateway when we get it all tuned up. At that point we just switch the endpoint.

src/modules/server-side/getProfile.ts
```diff
@@ -88,3 +94,64 @@ export const getPropsForWalletOrUsername: GetServerSideProps<WalletDependantPage
...
+    bannerImageUrl: string;
+  } | null = null;
+  if (publicKey) {
+    const result = await getProfileInfoFromPubKey({ pubKey: publicKey.toBase58() });
+    profileInfo = result?.wallet?.profile ?? null;
+  } else if (isTwitterUsername(input)) {
+    const result = await getProfileInfoFromTwitterHandle({ handle: input });
+    profileInfo = result?.profile ?? null;
+    if (profileInfo?.walletAddress) {
+      publicKey = getPublicKey(profileInfo.walletAddress);
```
kespinola, line 126: We chatted at standup and think you are adjusting but I'm worried about page load performance with querying so much on the server. Server side queries should be reserved for data we need to decide on routing choices, denials, and meta head tag info.

kespinola (changes requested):
Its looking good but like to remove the 2 clients and query less from the server side props.

austinmilt (approved):

PR #310, opened 2022-04-05
Milo/feat/make offer nft

### LINKED ISSUE #294: Make an Offer on NFT
Screen Shot 2022-04-04 at 12 33 23 PM

```
/nfts/{address}/offers/new
```

```
query {
market(subdomain: "holaplex") {
  auctionHouse {
    ...
  }
}

nft(address: "") {
  offers {
    ...
  }
}
}
```

### References

https://github.com/holaplex/marketplace/blob/dev/src/pages/nfts/%5Baddress%5D.tsx#L706

https://github.com/holaplex/marketplace/blob/dev/src/components/Offer/index.tsx#L47

milorue:
1) Implemented new offer when not listed designs
 - Modal
 - Blur Page util layout
 - New Offer Form
 - Make Offer block on Nft page
2) Added form resolver in zod (added dependency 10.6KB minified) and react-hook-form
3) Skeleton work for continued work
4) Currently utilizing the program setup used in marketplaces once the SDK is ready will need to be refactored to use that

*Note:* I'm merging these into a big feature branch for ease of review
Closes #294

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,9 @@
     "@holaplex/js": "^4.3.2",
     "@holaplex/mpl-auction-house": "^0.0.3",
     "@holaplex/ui": "^0.0.31",
+    "@hookform/resolvers": "^2.8.8",
+    "@metaplex-foundation/mpl-auction-house": "^1.1.1",
+    "@metaplex-foundation/mpl-token-metadata": "^1.2.5",
     "@metaplex/js": "^4.2.0",
     "@prisma/client": "^3.9.2",
     "@project-serum/anchor": "^0.20.1",
@@ -84,7 +87,7 @@
     "react-color": "^2.19.3",
     "react-device-detect": "^1.17.0",
     "react-dom": "17.0.2",
-    "react-hook-form": "^7.17.1",
+    "react-hook-form": "^7.29.0",
     "react-infinite-scroll-hook": "^4.0.1",
     "react-intersection-observer": "^8.33.1",
     "react-loader-spinner": "^5.1.4",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -84,7 +87,7 @@
     "react-color": "^2.19.3",
     "react-device-detect": "^1.17.0",
     "react-dom": "17.0.2",
-    "react-hook-form": "^7.17.1",
+    "react-hook-form": "^7.29.0",
```
milorue, line 90: Might be a breaking change I'll look into this shortly

src/common/components/forms/OfferForm.tsx
```diff
@@ -0,0 +1,228 @@
+import React, { FC, useEffect, useState } from 'react';
```
kespinola, line 1: Going to offer some more refactor recommendations for the repo but propose we keep this flat at `src/components/OfferForm`

pages/nfts/[address]/offers/new.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  if (called && !data?.nft && !loading) {
+    return <Custom404 />;
+  }
+
+  return (
+    <>
+      <BlurPage>
+        <NftByAddress address={address} />
+      </BlurPage>
```
Comment (kespinola) [inline pages/nfts/[address]/offers/new.tsx:53]: nb: I think layout still worth a review. https://nextjs.org/docs/basic-features/layouts#per-page-layouts

Comment (kristianeboe) [reply pages/nfts/[address]/offers/new.tsx:53]: I suggest we go through the app after this MVP and define a few layouts we can implement like
- discovery
- profiles
- details

probably will need a few nested layouts as well.

For now, as far as I can "see with my eyes" React does not do undue renders if the Wrapper components are the same between pages so I don't think it's critical

kristianeboe (approved):
Code looks good, but I didn't actually test if it works 😅

PR #318, opened 2022-04-11
Profile Created Tab

### LINKED ISSUE #305: A created tab in Profiles
- [x] Get the created query in the frontend
```
query {
  nfts(creators: ["4QawWnjkTW91g3Qvh3cz39C9xQ4VQB6s181XB6xAe6hY"], limit: 10, offset: 0) {
    image
  }
}
```
- [x] Add a new tab next to owned called created
- [x]  Change owned to collected

milorue:
closes #305 

- Added created tab
- Updated `Owned` to `Collected`
- Added creator query to graphql-codegen

### DIFF
--- a/src/common/components/elements/ProfileMenu.tsx
+++ b/src/common/components/elements/ProfileMenu.tsx
@@ -10,6 +10,7 @@ import { useProfileData } from '@/common/context/ProfileData';
 
 enum TabRoute {
   OWNED,
+  CREATED,
   ACTIVITY,
 }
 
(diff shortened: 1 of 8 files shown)


src/common/components/elements/ProfileMenu.tsx
```diff
@@ -10,16 +10,32 @@ import { useProfileData } from '@/common/context/ProfileData';
...
+        setSelectedIndex(TabRoute.CREATED);
+        break;
+      case `/profiles/[publicKey]/activity`:
+        setSelectedIndex(TabRoute.ACTIVITY);
+        break;
+      default:
+        setSelectedIndex(TabRoute.OWNED);
+        break;
+    }
+  }, [pathName, path]);
```
kespinola, line 37: Our tabs should be based on the path and when we change tab that should change the route. Doing as an effect with set state leads to unknown conditions on the initial render.

kespinola (changes requested):

src/common/components/elements/ProfileMenu.tsx
```diff
@@ -34,7 +50,21 @@ export const ProfileMenu: FC = () => {
                 )}
               >
                 <FeatherIcon height={16} width={16} icon="image" className="mr-4" />
-                Owned
+                Collected
+              </button>
+            )}
+          </Tab>
+          <Tab as={Fragment}>
```
kespinola, line 57: Ok so tabs group works with indexes. In this case want something like

```
['/profiles/[publicKey]/created', '/profiles/[publicKey]/activity', '/profiles/[publicKey]'].indexOf(router.path)
```

src/common/components/elements/ProfileMenu.tsx
kristianeboe: @kespinola the  tabs are based on the Routes, this is just to apply "selected" styling to the correct tab

PR #331, opened 2022-04-15
#323: Homepage v2 Footer

### LINKED ISSUE #323: New Footer
image

To get your feet wet hehe :) 

Create this footer as a component 
- You can get the social links and icons from the current Hero section
- Use tailwind for everything
- make it responsive (easy using tailwind's sm:, md: etc. selectors

Shout out if you have any questions

austinmilt:
This PR adds a new `Footer` component to be used (primarily) in the v2 homepage. I have tentatively used it in a placeholder copy of `index.tsx` (`index2.tsx`).

closes #323


src/common/components/home/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+                            <li><FooterLink href="/marketplace/new"><TaggedNew>Marketplaces</TaggedNew></FooterLink></li>
+                            <li><FooterLink href="/storefront/new">Set up a storefront</FooterLink></li>
+                        </ul>
+                    </div>
+                    <div className="justify-start">
+                        <span className="text-white text-base font-medium">Community</span>
+                        <ul className="mt-4 space-y-2">
+                            <li><FooterLink newTab href="https://github.com/holaplex">GitHub Repos</FooterLink></li>
+                            <li><FooterLink newTab href="https://airtable.com/shrXoZ9qU9g4oeM81">Hola Collective</FooterLink></li>
+                            <li><FooterLink href="#">HolaListings</FooterLink></li>
```
austinmilt, line 41: note placeholder link

src/common/components/home/Footer.tsx
```diff
@@ -0,0 +1,75 @@
...
+    );
+}
+
+
+interface LinkProps {
+    href: string;
+    newTab?: boolean;
+}
+
+const FooterLink: FC<LinkProps> = props => {
```
austinmilt, line 64: does my FooterLink class makes sense, as it essentially just wraps up some tailwind classes with no useful additional logic, which seems like a tailwind no-no but also makes my life easier

src/common/components/home/Footer.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { FC, VFC } from 'react';
+import SocialLinks from '../elements/SocialLinks';
+
+const Footer: VFC = () => {
+    return (
+        <footer className="bg-gray-800 p-10">
+            <div className="flex flex-row flex-wrap justify-between">
```
austinmilt, line 7: is this combo of flex and grid (see nested elements) good practice?

milorue: yeah its fine to have a parent flex element with a grid child

src/common/components/home/Footer.tsx
milorue: @austinmilt this is good you are just setting the default for children elements could be overridden it on each child element

milorue (approved):

PR #373, opened 2022-04-22
Add cardinal link profile button

echohtp:
Before 
image

After
image


dandlezzz:
How does the behave if the user has connected their profile to cardinal-twitter already, how does it behave if they have connected to bonfida ?

echohtp:
> How does the behave if the user has connected their profile to cardinal-twitter already, how does it behave if they have connected to bonfida ?

@dandlezzz 

there is no differing state btwn connected and disconnected - not until we've indexed their data and have it available on our API.

dandlezzz (approved):
bring on the cardinal

Issue #334, opened 2022-04-18
Homepage v2: Marketplaces section


austinmilt:
Create the Marketplaces preview section
![image](https://user-images.githubusercontent.com/1540936/163832030-5b1b4aaa-2043-4955-a931-b699f93b925a.png)

- preview 4 marketplaces in carousel
- carousel should have right/left arrows to see next/previous page with a slide transition
- carousel navigation arrows should grow on hover
- carousel navigation arrows should only appear as long as there is content in that direction
- link to waitlist in `Join Waitlist` (https://marketplace.holaplex.com/waitlist)

austinmilt:
@bellebellesol @damianr 
please review

bellebellesol:
this looks good. one thing missing from the design for these cards that i'll need to do is to differentiate which marketplaces are live and which are coming soon (these won't link to anything so they'll be in disabled type state).

bellebellesol:
supplement creator PFPs with a 'coming soon' chip on marketplaces cards that are not yet live. remove hover interactions and drop shadows from these cards to communicate that they're not clickable. also, make the overlay 80% opacity instead of the gradient to darken and give further visual separation. what do you think about this @damianr? open to any ideas for this. 

<img width="321" alt="Screen Shot 2022-04-18 at 5 36 49 pm" src="https://user-images.githubusercontent.com/95664179/163841194-39f932a0-37d4-4915-aaf9-f75c9987772b.png">
<img width="668" alt="Screen Shot 2022-04-18 at 5 36 34 pm" src="https://user-images.githubusercontent.com/95664179/163841199-d8794a70-a4fc-4899-9f98-e581a10a1548.png">

austinmilt:
Additional Notes/Requirements

## Marketplaces

### Data

- list of featured marketplaces (as subdomain string)
- other data already taken care of

### Unmet Data Needs

- gql endpoint for list of featured marketplaces (as subdomain string), with limit
    - items listed
    - floor price
    - average sale price

austinmilt:
@kespinola the Graph QL query would look something like

```graphql
query FeaturedMarketplacesQuery($limit: Int = 10, $offset: Int = 0){
  featured_marketplaces(limit: $limit, offset: $offset) {
    marketplace{...}
  }
}
```

where `marketplace{...}` is the existing marketplace schema

PR #390, opened 2022-04-26
#334: Featured Marketplaces Section

austinmilt:
This PR adds a preview for featured marketplaces to a WIP v2 homepage.

# Desktop Styling
image

# Mobile Styling
image

Start in `FeaturedMarketplacesSection`

# Major Changes
- added WIP v2 homepage at route `/home-v2-wip` (`home-v2-wip.tsx`)
- added components for consistent formatting of home sections, including a compound component for the home section layout (`HomeSection.tsx`) and a link  component that differentiates between internal and external routes/links (`HomeLink.tsx`)
- added featured marketplaces section component, which has a hardcoded list of marketplaces to feature, most of which will be removed before adding to the final new page (`FeaturedMarketplacesSection.tsx`, `home-v2-wip.tsx`, `index.d.ts`, `yarn.lock`)
- updated marketplace preview component to handle missing auctionhouse stats and banner URL (`MarketplacePreview.tsx`, `types.d.ts`)

# Other Changes
- removed some unused imports (`ActivityContext.tsx`, `AppHeader.tsx`, `WalletIndicator.tsx`)
- refactored some image utils for getting fallback images to reduce duplicate code and work when a pubkey is in the wrong format (`image.ts`)
- made `ChevronRight` easier to format externally


src/common/components/home/MarketplacePreview.tsx
```diff
@@ -15,45 +17,66 @@ const LoadingPreview = () => {
...
+          target="_blank"
+          rel="noreferrer"
+          onClick={onClickMarketplaceLink}
+        >
           <img
-            src={imgOpt(data.bannerUrl, 600)}
+            src={imgOpt(data.bannerUrl, 800)}
             alt={`${data.name}`}
-            className="absolute w-full object-cover"
+            className="min-h-full min-w-full object-cover"
```
kristianeboe, line 73: should this be h-full and w-full?

kristianeboe (commented):
Looks pretty good to me! A bit heavy handed with typings in some places (I prefer to rely a bit more on the implicit typings instead to keep the code more concise), but I have no problem with it.

Good work mate!

src/common/components/home/MarketplacePreview.tsx
austinmilt: No. The banner aspect ratio doesnt match that of the marketplace preview container. We want to force it to cover the whole thing (with `object-cover`) but to not deform the banner image, thus using (`min-h-full min-w-full`)

austinmilt:
Made changes from @bellebellesol feedback, primarily to do with avatars. See examples below

# Desktop
image


# Mobile
image

austinmilt:
> Looks pretty good to me! A bit heavy handed with typings in some places (I prefer to rely a bit more on the implicit typings instead to keep the code more concise), but I have no problem with it.
> 
> Good work mate!

Thanks!

I opt for explicit typing everywhere that it's not extremely obvious. I find it makes the code easier to understand because you dont have to guess what the types of variables are, which is great when you dont have the luxury of the IDE to infer (e.g. during PR review and reading open-source code on gh)

pages/home-v2-wip.tsx
```diff
@@ -0,0 +1,16 @@
...
+    <>
+      <div className="container mx-auto pb-48 w-[80%]">
+        <FeaturedMarkeplacesSection />
+      </div>
+      <Footer />
+    </>
+  );
+};
+
+export default Home;
```
kespinola, line 16: Why even have a page component if its going to import everything from another component?

Also personally don't like things like HomeSection, HomeSection.Body. What I do like is shared layouts objects like Section, Row, Col.

Tailwind is easy enough to work with just use vanilla html elements. This is 17 files that could really be like 5. 

Not trashing on it just want to get my concerns across.

src/common/components/elements/Avatar.tsx
```diff
@@ -10,10 +10,23 @@ import { useWallet } from '@solana/wallet-adapter-react';
 
 export const AvatarIcons = ({ creators }: { creators: { address: string }[] }) => {
   return (
-    <div className="relative">
-      {creators.map(({ address }, i) => (
-        <AvatarIcon address={address} index={i} key={address} />
+    // wrap the avatars in a container with a small, colored background
+    <div className={`inline-flex items-center rounded-full bg-gray-600 bg-opacity-70 p-1`}>
+      {creators.slice(0, 4).map(({ address }, i) => (
+        <div key={address} className={`h-6 w-6 ${i > 0 ? '-ml-3' : ''}`}>
```
kespinola, line 16: Check out classnames package for doing conditional classnames.

https://www.npmjs.com/package/classnames

Already included in the project.

pages/home-v2-wip.tsx
austinmilt: > Why even have a page component if its going to import everything from another component?

@kristianeboe is right that this is a placeholder for the future v2 `index.tsx`

> ChevronRight could also be replaced entirely everywhere with an export from Heroicons/outline.

I believe I'm using the hero icons throughout my code. I will create a ticket for changing usages everywhere, though (#406).

> As for an explosion in number of components, maybe we can find a middle ground? IMO we should move code from HomeSection and HomePage back into the home-v2-wip page, but keep FeaturedMarketplacesSection and MarketplacePreview(Card).

Will do this. 

I disagree with the tailwind approach that everything should be flat and you should avoid wrapping stuff into components. While it's true I have wound up with some specialized component names, it's also true that repeated giant classNames and nested elements are hard to understand and maintain, given how complex and fragile markup is. I shouldnt need to check every `div` at the same level for a `hover` selector to know whether or not something happens when it is hovered, and that every `div` has the same hover behavior.

austinmilt: Actually @kristianeboe if I want to get rid of `HomeSection` , all of that will move into `FeaturedMarketplacesSection`. I will then have to repeat the same markup in `FeaturedBuyNow`, `FeaturedProfiles`, etc. So, if we are getting rid of `HomeSection`, then IMO the whole thing should flatten into `home-v2-wip` as @kespinola suggests, so that the shared code can be seen in the same document.

austinmilt: I moved everything into `home-v2-wip`, but also repeated some stuff in `HomeV1FeaturedMarketplacesSection` so I could include that (temporarily) in the existing homepage

PR #415, opened 2022-05-04
try twitter profile link first and fall back to shortened address

richardfsr:


dandlezzz:
@richardfsr thanks for this!

PR #475, opened 2022-05-20
Update the about page

CalebBabin:


CalebBabin:
Use `kebo/test/circleclick` instead of `main`

PR #487, opened 2022-05-23
homepage v2: dynamic carousel columns

austinmilt:
sets the number of columns in homepage carousels based on window size, which helps avoid the issue of some cards being differently sized from others within a carousel


milorue (approved):
Lgtm

PR #509, opened 2022-05-31
Kris/feat/debounce and track search

kristianeboe:
Adds
- Debounce for search with min 2 chars and 300 ms to reduce API calls (and tracking calls)
- Analytics events for search queries and result clicks

### DIFF
--- a/src/common/components/search/SearchBar.tsx
+++ b/src/common/components/search/SearchBar.tsx
@@ -40,6 +40,7 @@ export const isPublicKey = (address: string) => {
 
 const SearchBar: FC<SearchBarProps> = ({ shortcut }) => {
   const searchResultsRef = useRef<HTMLDivElement>(null!);
+  const searchInputRef = useRef<HTMLInputElement>(null!);
 
   const router = useRouter();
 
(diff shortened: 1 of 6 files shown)


src/common/components/search/SearchBar.tsx
```diff
@@ -120,26 +146,29 @@ const SearchBar: FC<SearchBarProps> = ({ shortcut }) => {
...
               </span>
 
-              <input
+              <DebounceInput
+                minLength={2}
+                debounceTimeout={300}
                 id="search"
                 autoComplete={`off`}
                 autoCorrect={`off`}
-                className="block w-full rounded-full border border-transparent bg-transparent py-3 pl-12 pr-2 text-base placeholder-transparent transition-all focus:border-white focus:placeholder-gray-500 focus:outline-none focus:ring-white sm:text-sm"
```
milorue, line 133: Reason for more padding?

milorue (approved):

src/common/components/search/SearchBar.tsx
kristianeboe: it was a bit off with the hover effect of the search icon. I'll perfect it when I look at keyboard nav in the future

PR #512, opened 2022-06-01
Milo/feat/consolidated feed events

milorue:
- Added follow aggregation card
- Added sale/nft event aggregation card
  - Added nft carousel component
- Updated how we aggregate events to be a bit more performant and easier to understand

*Note:* I left some code commented out as I want to return to it but it isn't needed in this pr and requires some further design discussion on how consolidated we want events to be


src/common/components/feed/feed.utils.ts
```diff
@@ -138,19 +147,83 @@ export function generateFeedCardAttributes(
...
+        profile: user.profile,
+      });
+    } else {
+      // do nothing
+    }
+  });
+  return users as User[];
+};
+
+export const aggregateEventsTime = (events: FeedQueryEvent[]) => {
```
austinmilt, line 169: The choice of averaging the event times seems a little odd to me. If they are all very close in time then it doesnt make much difference, but if they are several minutes apart, I would opt for taking the earliest time to indicate "This set of events started at this time".  Kinda like if you went to a concert and averaged the start time of the cover band with the main show, versus saying the show started when the cover band did.

src/common/components/feed/feed.utils.ts
```diff
@@ -134,19 +143,83 @@ export function generateFeedCardAttributes(
...
+  events.forEach((event) => {
+    const date = DateTime.fromISO(event.createdAt).toMillis();
+    avgTime += date;
+  });
+  avgTime = avgTime / events.length;
+  const avgDate = DateTime.fromMillis(avgTime);
+  return avgDate;
+};
+
+export function shouldAggregateFollows(e1: FeedQueryEvent, e2: FeedQueryEvent, e3: FeedQueryEvent) {
```
austinmilt, line 180: It's not clear to me why this takes 3 events rather than 2, or 5, or N. What are the consequences of passing in an arbitrarily sized array of events?

src/common/components/feed/feed.utils.ts
```diff
@@ -134,19 +143,83 @@ export function generateFeedCardAttributes(
...
-    e2.walletAddress === e3.walletAddress
+    // e1.walletAddress === e2.walletAddress &&
+    // e2.walletAddress === e3.walletAddress &&
+    e1.__typename === 'FollowEvent' &&
+    e2.__typename === 'FollowEvent' &&
+    e3.__typename === 'FollowEvent'
   );
 }
 
+export const shouldAggregateSaleEvents = (
```
austinmilt, line 196: Same comment as above. Why not any number of events?

src/common/hooks/useFocused.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react';
```
austinmilt, line 1: I dont see this being used anywhere. Am I missing it?

src/common/hooks/useFocused.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const useFocused = () => {
+  const [focused, setFocused] = useState(document.activeElement);
```
austinmilt, line 4: fwiw I ran into an issue the other day trying to use `document` into a `useMemo` hook and it failed because `document` did not exist yet. AFAICT this has potential to break as well unless you always use this in a `useEffect` or put the `useEffect` in this hook. Or iirc you can do a check with `if (typeof window !== 'undefined')  //then document will exist`

pages/alpha/index.tsx
```diff
@@ -192,13 +189,16 @@ const AlphaPage = ({ address }: { address: string }) => {
...
+      if (shouldAggregateFollows(event, nextEvent, nextNextEvent)) {
+        // remove the item that would start the aggregation
+        feedItems.pop();
         // const eventsAggregated: FeedQueryEvent[] = feedEvents
         //   .slice(i)
         //   .filter((fe, i, arr) => shouldAggregate(fe, arr[i + 1]));
         const eventsAggregated: FeedQueryEvent[] = [feedEvents[i]];
         let j = i + 1;
-        while (shouldAggregate(feedEvents[j], feedEvents[j + 1], feedEvents[j + 2])) {
+        while (shouldAggregateFollows(feedEvents[j], feedEvents[j + 1], feedEvents[j + 2])) {
```
austinmilt, line 201: Are `j`, `j+1` and `j+2` guaranteed to exist? I see in already-committed-code above that there's references to `i+1` and `i+2`, but it's not clear to me those are safe either. And it seems... dirty to try to reference them if they dont.

src/common/components/feed/FeedCard.tsx
```diff
@@ -412,6 +479,30 @@ const OfferAction = (props: { nft: any }) => {
...
+        <div
+          key={user.address}
+          className={`${
+            i > 0 && `-ml-6`
+          } transition duration-100 ease-in hover:z-10 hover:scale-105`}
+        >
+          <ProfilePFP user={user} />
+        </div>
+      ))}
+      {users.length - 3 > 1 && (
```
austinmilt, line 495: would be clearer as `users.length > 4` or, if you want to keep the `3` in there, then extract it to a constant and use it everywhere... here I would use it like `users.length > PFP_STACK_MORE_THRESHOLD + 1`

Also, should it be `users.length > 3`?

austinmilt (commented):
I see a couple potential bugs, but not sure. Some other feedback as well

src/common/components/feed/feed.utils.ts
milorue: should only aggregate after 3

milorue: but yes I think thats a good change to be able to accept n

src/common/hooks/useFocused.ts
milorue: No longer used will delete

src/common/components/feed/feed.utils.ts
milorue: Hard coded to 3 for now since we only want to aggregate at 3 or above

src/common/components/feed/FeedCard.tsx
milorue: good suggestion fixed

PR #519, opened 2022-06-03
Collections POC + various ways of doing it (layouts vs dynamic urls etc.)

kristianeboe:
Here is another thing to look at

It shows 3 different ways of laying out collections under the folders pages/collections, pages/collections2 and pages/collections3
- collections is a double dynamic nested page with [collectionAddress] and [collectionTab] and it uses the collectionTab determined by ssr to show which tab is active
- collections2 is the one I showed you yesterday that relies more on headlessUi to keep the tabs in state, while updating url after each tab change. I think this one performs far better locally and is the cleanest.
- collections3 is an attempt to use Nextjs layouts. The files do get a bit smaller, but are in fact a bit harder to reason about.

There is also a folder called /collections under elements with a util file and a few smaller components.

Please take a look and see if you have any preferences between the ways. You can also check it out locally to see performance, and even on .dev unless someone removes it. You might have to reload a few times to get them to work as there are still some bugs.

https://www.holaplex.dev/collections/DTjENPCxVnukLh5wbdBsb3CDttSfQzdNeRyk7dzr6vjr/nfts
https://www.holaplex.dev/collections2/DTjENPCxVnukLh5wbdBsb3CDttSfQzdNeRyk7dzr6vjr/nfts
https://www.holaplex.dev/collections3/DTjENPCxVnukLh5wbdBsb3CDttSfQzdNeRyk7dzr6vjr/nfts


src/graphql/collections/nftCollection.graphql
```diff
@@ -0,0 +1,105 @@
+query nftCollection(
```
austinmilt, line 1: This query is massive. Do we need all these data?

src/graphql/collections/nftsInCollection.graphql
```diff
@@ -0,0 +1,97 @@
+query nftsInCollection(
```
austinmilt, line 1: This query is massive. Do we need all these data?

src/common/components/collections/CollectionPreviewCard.tsx
```diff
@@ -0,0 +1,29 @@
+import Link from 'next/link';
+import React from 'react';
+
+interface CollectionCardProps {
+  collectionAddress: string;
```
austinmilt, line 5: just a tiny comment: you shouldnt prefix everything with `collection` inside an interface called `CollectionCardProps`. Better for the default to be that it's part of the collection, and to prefix anything that isnt (e.g. `ownerAddress`)

src/layouts/CollectionLayout.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <ProfileHandle user={props.user} />
+    </span>
+  );
+}
+
+export default function CollectionLayout({
+  children,
+  ...props
+}: {
+  collectionAddress: string;
```
austinmilt, line 36: I think here you could use `: CollectionPage & {children: ReactElement}` instead of having to repeat the interface spec

src/layouts/CollectionLayout.tsx
```diff
@@ -0,0 +1,183 @@
...
+            selected={router.pathname.includes('nfts')}
+            url={`/collections3/${props.collectionAddress}/nfts`}
+          />
+          <Tab1
+            title={'About'}
+            selected={router.pathname.includes('about')}
+            url={`/collections3/${props.collectionAddress}/about`}
+          />
+        </div>
+        <div className="">
```
austinmilt, line 159: empty `className`

src/layouts/CollectionLayout.tsx
```diff
@@ -0,0 +1,183 @@
...
+            collectionFollowers,
+            creators,
+          })}
+        </div>
+      </div>
+    </div>
+  );
+}
+
+const Tab1 = (props: { url: string; selected: boolean; title: string }) => (
```
austinmilt, line 171: Why not `Tab`?

austinmilt (commented):
I like the layout approach compared to the other two because

1. Pages can be more concise
2. Shared data fetching takes place in the shared layout (DRY)
3. Uses features out of the box from Nextjs

For me, the main source of confusion is in how pieces are hooked together, and where some data _magically_ come from. However, once I understood how the `.getLayout` method worked, it became much clearer.

pages/collections3/[collectionAddress]/nfts.tsx
```diff
@@ -0,0 +1,99 @@
+import { CollectionRaisedCard } from '@/common/components/collections/CollectionRaisedCard';
```
Comment (kespinola) [inline pages/collections3/[collectionAddress]/nfts.tsx:1]: I really like the layouts here. easy to understand and I feel like I can jump in.

pages/collections3/[collectionAddress]/about.tsx
```diff
@@ -0,0 +1,59 @@
...
+              }}
+            />
+          ))}
+        </div>
+      </CollectionRaisedCard>
+    </div>
+  );
+}
+
+CollectionAboutPage.getLayout = function getLayout(page: ReactElement) {
```
Comment (kespinola) [inline pages/collections3/[collectionAddress]/about.tsx:57]: I recommend doing the layout as a component like in marketplace so it has access to the app context and can get server side props.

https://github.com/holaplex/marketplace/blob/dev/src/pages/_app.tsx#L81

src/common/components/collections/CollectionPreviewCard.tsx
kristianeboe: In general I agree, but because we do a lot of destructuring in various places it quite often becomes clearer when you have the prefix! Of course, we can just rename it in those instances. I'll take a note to refactor these soon :)

pages/collections3/[collectionAddress]/about.tsx
Comment (kristianeboe) [reply pages/collections3/[collectionAddress]/about.tsx:57]: This is clever!

Comment (austinmilt) [reply pages/collections3/[collectionAddress]/about.tsx:57]: @kristianeboe Copying marketplaces exactly made typescript unhappy. I have made updates in [my branch](https://github.com/holaplex/holaplex/compare/austin/348/discover?expand=1) that seem to do the trick (see _app.tsx)

PR #535, opened 2022-06-13
fix: refactored User interface

kristianeboe:


src/common/components/feed/FeedCard.tsx
```diff
@@ -554,7 +555,7 @@ export function ProfilePFP({ user }: { user: User }) {
       <a target="_blank">
         <img
           className={classNames('rounded-full', 'h-10 w-10')}
-          src={user?.profile?.profileImageUrl || getPFPFromPublicKey(user.address)}
+          src={user?.profile?.profileImageUrlLowres || getPFPFromPublicKey(user.address)}
```
austinmilt, line 558: should you use `??` here?

src/common/components/feed/feed.utils.ts
```diff
@@ -7,23 +7,27 @@ import {
   ListingReceipt,
   MintEvent,
   PurchaseReceipt,
+  WhoToFollowQuery,
 } from 'src/graphql/indexerTypes';
 
 type FeedEventTypes = FeedItem['__typename'];
 export type FeedQueryEvent = FeedQuery['feedEvents'][0];
+export type User = WhoToFollowQuery['followWallets'][0];
```
austinmilt, line 15: This feels very awkward to me, to use an externally defined data structure to define what is in truth an internal representation of data needed for an internally used object. While I still contend that it's better to define the interface for each component and _transform_ external data into that format, a compromise solution would be to define the internal data interface without respect to the external data interface but such that it matches the external one, and then cast the external to the internal. e.g.

```
export type User = {
   address: string;
   profile: {
      handle: string;
      profileImageUrlLowres: string;
   }
}
```

And then below, you can still do `{<...graph-ql data>} as User`

In this way, the internal representation is the authority, and we cast the external data to this format. One side benefit is that we can more tightly control what minimum data are required to construct/display a component, and more quickly catch issues when the external data structure is lacking. This is a pretty common bug in Holaplex.com currently that arises out of the approach your PR continues to use.

PR #538, opened 2022-06-13
featured profiles fetch poc

austinmilt:
This PR shows how we can leverage hooks to simultaneously

1. execute queries at the page level 
2. allow components to define their own data interfaces
3. abstract away API data structures and client logic
4. keep components narrowly scoped

Notice:
- how much code I was able to delete from the component files when I didnt have to worry about whether some graph-ql data were present or not (because the component's interface dictates what's valid data to pass in) or how/when to update state. 
- how much flatter the data interfaces could become when the inherent nested structure of the Graph QL objects was removed.
- that the outer `FeaturedProfilesSection` component didnt need to pass down, transform, or even know about the data structure of the `ProfilePreview` card. It only has to pass along the key and let the `ProfilePreview` get its own data.

This is not a complete solution (yet), as we should explore how to cache the transformed data rather than re-transform each time the hook is called. We would also want to expose some of the other functionality in Apollo, like refetching, updating, etc.

### DIFF
--- a/src/common/components/home/FeaturedProfilesSection.tsx
+++ b/src/common/components/home/FeaturedProfilesSection.tsx
@@ -1,57 +1,26 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { HomeSection, HomeSectionCarousel } from 'pages/index';
-import { ProfilePreviewData } from '@/types/types';
-import {
-  useFeaturedProfilesQuery,
-} from 'src/graphql/indexerTypes';
-import {
-  useWallet,
-  WalletContextState,
-} from '@solana/wallet-adapter-react';
 import ProfilePreview from '../elements/ProfilePreviewCard';
+import { QueryContext, useFeaturedProfiles } from '@/common/hooks/home';
 
 const CAROUSEL_ROWS: number = 2;
 const CAROUSEL_COLS: number = 3;
-const CAROUSEL_PAGES: number = 2;
-const N_LISTINGS: number = CAROUSEL_ROWS * CAROUSEL_COLS * CAROUSEL_PAGES;
-
-//TODO remove once other profiles have enough followers to preclude this one in the backend
-const DISALLOWED_PROFILES: string[] = ['ho1aVYd4TDWCi1pMqFvboPPc3J13e4LgWkWzGJpPJty'];
 
 export default function FeaturedProfilesSection(): JSX.Element {
-  const wallet: WalletContextState = useWallet();
-  // initial value hack to get loading card
-  // TODO pass in a loading boolean to bypass loading until the address is known
-  const [featuredProfiles, setFeaturedProfiles] = useState<ProfilePreviewData[]>(
-    [...Array(N_LISTINGS)].map((_) => ({ address: '', profile: {}, nftCounts: {} }))
-  );
-
-  const dataQuery = useFeaturedProfilesQuery({
-    variables: {
-      userWallet: wallet?.publicKey,
-      limit: CAROUSEL_PAGES * CAROUSEL_COLS * CAROUSEL_ROWS,
-    },
-  });
-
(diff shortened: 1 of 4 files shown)


src/common/hooks/home.ts
```diff
@@ -0,0 +1,84 @@
...
+//TODO remove once other profiles have enough followers to preclude this one in the backend
+const DISALLOWED_PROFILES: string[] = ['ho1aVYd4TDWCi1pMqFvboPPc3J13e4LgWkWzGJpPJty'];
+
+function useFeaturedProfilesData() {
+  const wallet: WalletContextState = useWallet();
+  const [featuredProfilesQuery, queryContext] = useFeaturedProfilesLazyQuery();
+
+  useEffect(() => {
+    async function query(): Promise<void> {
+      await featuredProfilesQuery({ variables: { limit: 25, userWallet: wallet?.publicKey } });
```
kespinola, line 21: This going to lead to excessive querying where every component using this hook is going to make a request. Please give fragments a try to allow the FeaturedProfileSection to define its data needs but still have the query executed at the page level.

src/common/components/home/FeaturedProfilesSection.tsx
```diff
@@ -1,58 +1,27 @@
...
-      setFeaturedProfiles(featuredProfiles.filter((p) => p.address !== profileAddress));
+      excludedProfiles.push(profileAddress);
+      setExcludedProfiles(excludedProfiles);
     },
-    [featuredProfiles]
+    [excludedProfiles]
   );
 
+  const featuredProfiles: string[] = (queryContext.data ?? []).filter(a => !excludedProfiles.includes(a));
+
```
kespinola, line 24: Shouldnt we do the exclude at the api level so the limit still works. If you request 10 from the api but filter out 1 on the client you are going to have 9 objects.

kespinola (changes requested):

src/common/hooks/home.ts
```diff
@@ -0,0 +1,84 @@
...
+    queryContext.data?.followWallets.find((w) => w.address === address);
+
+  let error: Error | undefined = queryContext.error;
+  if (!error && gqlObject) {
+    if (gqlObject.nftCounts.created === null || gqlObject.nftCounts.created === undefined) {
+      error = new Error('Missing count of NFTs created.');
+    } else if (gqlObject.nftCounts.owned === null || gqlObject.nftCounts.owned === undefined) {
+      error = new Error('Missing count of NFTs collected.');
+    }
+  }
```
kespinola, line 65: If a fragment is used don't need to throw errors as the component says what data it needs and doesn't need to check.

src/common/components/home/FeaturedProfilesSection.tsx
austinmilt: I think this particular filtering should happen on the client side, though that could feed back into the query. The client needs to determine what data are required to display a profile card, whereas the backend needs to serve whatever data are requested.

I could add optional filters to every API to exclude specific keys - in this case wallet addresses excluded from follow_wallets - and let the frontend update the query after determining a profile isnt worth showing.

src/common/hooks/home.ts
austinmilt: I dont think fragments fix this issue.

Any data which the backend may optionally return will be potentially nullish on the frontend and so the frontend will may still need to check nullity of required data.

In this case, the Graph QL API shows that the NFT counts are required, so it doesnt apply from the perspective of the backend.

However, the Apollo generated types make pretty much everything nullable, meaning that the frontend still doesnt have a guarantee that the data will be there. So, in fact this check is still necessary.

src/common/hooks/home.ts
austinmilt: This is complementary to fragments, which I definitely want to use in addition to something like this.

We should be able to leverage apollo caching the query to avoid unnecessary triggers. However, there is probably some work to be done to get it working right.

PR #542, opened 2022-06-15
Updates to integrate new marketplace sdk

alchemistgo87:


austinmilt (commented):
Mostly looks fine. I dont understand the choice of auction houses, though.

src/common/components/forms/AcceptOfferForm.tsx
```diff
@@ -63,9 +69,15 @@ const AcceptOfferForm: FC<AcceptOfferFormProps> = ({
   const onAcceptOffer = async () => {
     if (offer) {
       if (listing) {
-        await sdk.offers(marketplace.auctionHouse).accept({ offer, nft, cancel: [listing] });
+        await sdk
+          .transaction()
+          .add(sdk.offers(marketplace.auctionHouses[0]).accept({ offer, nft, cancel: [listing] }))
```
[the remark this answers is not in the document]
alchemistgo87: Hi Austin, Right now holaplex is going to support just a single spl-token i.e sol. So, only one auctionHouse will be present in the list. We can modify this (as you suggested) as the requirement arises. 

And for listing there can be only one auctionHouse per listing as a particular listing can only belong to a single spl-token. Similarly for offer and purchase.

@kespinola Do you want to add something here?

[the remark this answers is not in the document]
alchemistgo87: That's correct! Apart from cases where we are creating a new offer/listing/purchase, it would be better to use auctionHouse directly from that particular object. 

And in cases where we are creating a new offer/listing/purchase we can get auctionHouse from marketplace.auctionHouses, example:

```
await sdk
        .transaction()
        .add(
          sdk
            .listings(
              marketplace.auctionHouses.filter(
                (ah) => ah.treasuryMint === selectedToken?.address
              )[0]
            )
            .post({
              amount,
              nft,
            })
        )
        .send()
```

kristianeboe:
We are still waiting for an update to the indexer here :) Just FYI

src/common/components/forms/CancelSellForm.tsx
```diff
@@ -51,7 +51,7 @@ const CancelSellForm: FC<CancelSellFormProps> = ({
       toast(`Canceling listing for ${nft.name}`);
       await sdk
         .transaction()
-        .add(sdk.listings(listing.auctionHouse).cancel({ listing, nft }))
+        .add(sdk.listings(listing.auctionHouse!).cancel({ listing, nft }))
```
kespinola, line 54: Why the `!`?

kespinola (changes requested):
Still has the merge conflicts.

src/common/components/forms/CancelSellForm.tsx
alchemistgo87: It is because in AhListing type auctionHouse is optional. So, here we were getting error that auctionHouse might be undefined.

kristianeboe: Is it easier to add a if(listing.auctionHouse) above?

alchemistgo87: Yes, that's possible too. Let me do that.

alchemistgo87: But we also have another option to remove optional from the auctionHouse, as in all listings/offers/purchases we will definitely have the auctionHouse. @kespinola should we remove optional for auctionHouse in the indexer?

src/common/components/forms/BuyForm.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { FC, useContext, useMemo } from 'react';
```
kespinola, line 1: File that needs to move based on new folder structure.

src/common/components/home/FeaturedBuyNowListingsSection.tsx
```diff
@@ -0,0 +1,184 @@
+import { useCallback, useEffect, useMemo, useState, VFC } from 'react';
```
kespinola, line 1: File that needs to move based on new folder structure.

kespinola (changes requested):
@kristianeboe can you prioritize moving the 2 files I noted to their new home in the folder structure?

Good opportunity to bring @alchemistgo87 up to speed on the folder structure.

kespinola (approved):

PR #548, opened 2022-06-21
Token Gated Admin Page

zacyungblut:
This PR adds a function to the path /admin/login that will display secret content only if the user holds a specific SPL token in his or her wallet.

### DIFF
--- a/pages/admin/login.tsx
+++ b/pages/admin/login.tsx
@@ -1,7 +1,4 @@
-import { useRouter } from 'next/router';
 import React, { useState, useEffect } from 'react';
-//import LitJsSdk from 'lit-js-sdk';
-import jwt_decode from "jwt-decode";
 const LitJsSdk = require('lit-js-sdk');
 
 
--- a/pages/admin/login.tsx
+++ b/pages/admin/login.tsx
@@ -3,7 +3,6 @@ import jwt_decode from "jwt-decode";
 const LitJsSdk = require('lit-js-sdk');
 import { Spinner } from '@/common/components/elements/Spinner';
 
-//6855BL6aaQW9NaJUtbMqyYXnewjWdnSrQxdarKQ54fdD
 const accessControlConditions = [
   {
     method: "balanceOfToken",
@@ -17,12 +16,9 @@ const accessControlConditions = [
   },
 ];
 
-
-
-
 const resourceId = {
-    baseUrl: 'http://localhost:3000',
-    path: '/admin/login',
+    baseUrl: process.env.NEXT_PUBLIC_BASE_URL,
+    path: process.env.NEXT_PUBLIC_ADMIN_PATH,
     orgId: "none",
     role: "none",
     extraData: "none"
@@ -48,14 +44,12 @@ async function fetchToken() {
     await litNodeClient.connect();
     let jwt;
     try {
-      console.log('>>>>>> attempting to fetch signed token! <<<<<<');
       jwt = await litNodeClient.getSignedToken(signingArgs);
     } catch(err) {
       console.log('error fetching jwt', err);
     }
 
     if (jwt) {
-      console.log('signing token found', { jwt });
       return jwt;
     }
 }
@@ -65,22 +59,15 @@ const renderSecretAdminContent = () => {
   return(<h1>Secret Admin Content</h1>);
 }
 
-
-
 export default function AdminLogin() {
   const [isAdmin, setIsAdmin] = useState(false);
 
   const validateUser = async () => {
     const jwt = await fetchToken();
-    console.log(JSON.stringify(jwt_decode(jwt), null, 2));
     let newResponse = (JSON.stringify(jwt_decode(jwt), null, 2));
     if (newResponse) {
       setIsAdmin(true);
     }
-    // if (payload.baseUrl !== "http://localhost:3000" || payload.path !== "/admin/login" || payload.orgId !== "" || payload.role !== "" || payload.extraData !== "") {
-    //   console.log('Success!');
-    //   setIsAdmin(true);
-    //}
   }
 
   useEffect(() => {
--- a/pages/admin/login.tsx
+++ b/pages/admin/login.tsx
@@ -1,26 +1,31 @@
 import React, { useState, useEffect } from 'react';
+import jwt_decode from "jwt-decode";
 const LitJsSdk = require('lit-js-sdk');
+import { Spinner } from '@/common/components/elements/Spinner';
 
-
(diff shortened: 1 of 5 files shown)


pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useState, useEffect } from 'react';
+import jwt_decode from "jwt-decode";
+const LitJsSdk = require('lit-js-sdk');
+import { Spinner } from '@/common/components/elements/Spinner';
+
+const accessControlConditions = [
```
austinmilt, line 6: Can you add a type declaration to this? Especially given that [getSignedToken accepts a plain js object](https://lit-protocol.github.io/lit-js-sdk/api_docs_html/#litnodeclientgetsignedtoken), it's not clear what's required.

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+    chain: "solana",
+    returnValueTest: {
+      key: "$.amount",
+      comparator: ">",
+      value: "0",
+    },
+  },
+];
+
+const resourceId = {
```
austinmilt, line 19:
```suggestion
const RESOURCE_ID = {
```

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const solAuthSig = await LitJsSdk.checkAndSignAuthMessage({ chain: 'solana' });
+
+    const chain = 'solana';
+    const permanent = false;
+
+    const signingArgs = {
+      solRpcConditions: accessControlConditions,
+      chain,
+      authSig: solAuthSig,
+      resourceId
```
austinmilt, line 41:
```suggestion
      RESOURCE_ID
```

pages/admin/login.tsx
austinmilt, line 19: Also add a type declaration to this if possible.

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+    },
+  },
+];
+
+const resourceId = {
+    baseUrl: process.env.NEXT_PUBLIC_BASE_URL,
+    path: process.env.NEXT_PUBLIC_ADMIN_PATH,
+    orgId: "none",
+    role: "none",
+    extraData: "none"
```
austinmilt, line 24: Looks like the recommended values are empty strings according to [the docs](https://lit-protocol.github.io/lit-js-sdk/api_docs_html/#resourceid).
```suggestion
    orgId: "",
    role: "",
    extraData: ""
```

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+async function fetchToken() {
+
+    const litNodeClient = new LitJsSdk.LitNodeClient();
+
```
austinmilt, line 30:
```suggestion
```

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+const resourceId = {
+    baseUrl: process.env.NEXT_PUBLIC_BASE_URL,
+    path: process.env.NEXT_PUBLIC_ADMIN_PATH,
+    orgId: "none",
+    role: "none",
+    extraData: "none"
+}
+
+async function fetchToken() {
```
austinmilt, line 27: declare return type please 🙏

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const permanent = false;
+
+    const signingArgs = {
+      solRpcConditions: accessControlConditions,
+      chain,
+      authSig: solAuthSig,
+      resourceId
+    }
+
+    await litNodeClient.connect();
```
austinmilt, line 44: Is this guaranteed not to error? Given that you swallow the error to get the token below, may be worth doing something like that here as well (or just add to the try/catch below)

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+    let jwt;
+    try {
+      jwt = await litNodeClient.getSignedToken(signingArgs);
+    } catch(err) {
+      console.log('error fetching jwt', err);
+    }
+
+    if (jwt) {
+      return jwt;
+    }
```
austinmilt, line 54: IMO if swallowing errors, you should return _something_ (can be null/undefined if that's a valid return value)
```suggestion
    return jwt;
```

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+const renderSecretAdminContent = () => {
+
+  return(<h1>Secret Admin Content</h1>);
```
austinmilt, line 59:
```suggestion
  return (<h1>Secret Admin Content</h1>);
```

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+export default function AdminLogin() {
```
austinmilt, line 62:
```suggestion
export default function AdminLogin(): JSX.Element {
```

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+    if (newResponse) {
+      setIsAdmin(true);
+    }
+  }
+
+  useEffect(() => {
+
+    validateUser();
+
+  }, [])
```
austinmilt, line 77: I think more idiomatic and perhaps more efficient would be to move the validation function down to the useEffect

```suggestion
    useEffect(() => {
      const validateUser = async () => {
        const jwt = await fetchToken();
        let newResponse = (JSON.stringify(jwt_decode(jwt), null, 2));
        if (newResponse) {
          setIsAdmin(true);
        }
      }
      validateUser();
  }, [])
```

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return (
+      <div className="mt-20 flex flex-col items-center text-center">
+        {isAdmin===false ?
+          <>
+        <h1 className="text-3xl font-medium font-semi-bold">
+          Verifying that you hold at least one Admin Token...
+        </h1>
+        <Spinner />
+          </>
+        : (
```
austinmilt, line 88: fix indentation

pages/admin/login.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <>
+          <h1 className="text-3xl font-medium font-semi-bold">
+            Welcome to the Holaplex Admin Page!
+          </h1>
+            </>
+        )
```
austinmilt, line 94: fix indentation

austinmilt (changes requested):
lots of formatting and a few suggestion to keep things in-line with the rest of the repo

Issue #342, opened 2022-04-19
Create a reusable popover card for profile previews


kristianeboe:
Currently we are using the antd popover, but we are experiencing a bug where if you scroll the page the popover follows down instead of staying attatched to its trigger
![image](https://user-images.githubusercontent.com/10433154/163989314-e79c14ea-60dc-4d88-bc58-497f39e54093.png)

What we want is ideally something like https://www.radix-ui.com/docs/primitives/components/hover-card

If we make this component generic it should be usable in 
- nav menu dropdowns
- sharing menus and other hover menus
- most importantly the NEW Profile hover cards
![image](https://user-images.githubusercontent.com/10433154/163990955-b364d6f4-f937-4e6f-a360-065fec6762bf.png)

Suggested implementation
- Take a look at the Headless UI component for Popover https://headlessui.dev/react/popover
- and this issue to make it hoverable https://github.com/tailwindlabs/headlessui/discussions/425
- it also needs to receive a prop to determine if it is triggered by hover, click, or both

PR #581, opened 2022-07-08
Create Generic popover with show on click and show on hover features

alchemistgo87:
Addresses #342

**1. Created a new generic Component named Popover with the following features:**
- Show on click
- Show on hover (Need to pass 'isShowOnHover = true')

**2. Now using this widget for:**
- Profile Image Popover
- Popover for user handles in feeds
- Popover for who to follow user profiles


src/components/PopoverWidget.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useState } from 'react';
+import { usePopper } from 'react-popper';
+
+interface PopoverProps {
+  children: React.ReactElement;
+  content: React.ReactElement;
+  isShowOnHover?: boolean;
+}
+
+const PopoverWidget = ({ children, content, isShowOnHover = false }: PopoverProps) => {
```
kristianeboe, line 11: I think we can just call this Popover, or even HoverCard. That is what Antd and Radix does. Which do you prefer?

https://www.radix-ui.com/docs/primitives/components/hover-card

https://ant.design/components/popover/

kristianeboe (approved):
Take a look at the comment, otherwise great work! :D

src/components/PopoverWidget.tsx
alchemistgo87: Sure, we can name it just 'Popover'
Sounds simple and better.

PR #592, opened 2022-07-12
fix: Removes need for docker and nodemon

kristianeboe:
Removes
- nodemon from package.json
- all use of prisma and docker
- the entire depreciated wallet provider
- the /api/wallets api routes
- the /archive/home (old home page)

You can still
- edit old stores
- presumably create new ones
Only from the footer though

I tested it quite a bit on localhost just clicking around everywhere, both signed in and not, so I think it's safe :)


pages/marketplace/new.tsx
```diff
@@ -74,7 +73,10 @@ export default function New() {
   const [form] = Form.useForm();
   const [pendingAddress, setPendingAddress] = useState<string>();
   const { setVisible } = useWalletModal();
-  const { solana, wallet, looking } = useContext(WalletContext);
+  const solana = useWallet();
```
austinmilt, line 76: This variable name is pretty confusing IMO. Better would be `userWallet`

pages/marketplace/new.tsx
```diff
@@ -74,7 +73,10 @@ export default function New() {
...
+  const { publicKey } = solana;
+  const pubkey = publicKey?.toBase58();
```
austinmilt, line 78: similarly, `pubkey` is pretty ambiguous IMO. Better would be like `userWalletPubKey` or something

src/lib/utils/index.ts
```diff
@@ -15,8 +15,8 @@ export const maybeCDN = (uri: string) => {
 export const getEnvironmentVariable = (environmentVariable: string): string => {
   const unvalidatedEnvironmentVariable = process.env[environmentVariable];
   if (!unvalidatedEnvironmentVariable) {
-    return '';
     console.error(`Couldn't find env var: ${environmentVariable}`);
+    return '';
```
austinmilt, line 19: IME environment variables should be considered prerequisites for startup/functioning, so a missing environment variable should lead to complete failure, or at least throw an error that must be handled by the caller.

src/modules/storefront/provider.tsx
```diff
@@ -1,35 +1,27 @@
...
-  storefront?: Storefront;
-};
-
-type StorefrontProviderProps = {
-  wallet?: Wallet;
-  children: (props: StorefrontProviderChildrenProps) => React.ReactElement;
-};
-
-export const StorefrontProvider = ({ wallet, children }: StorefrontProviderProps) => {
+export const StorefrontProvider: FC = ({ children }) => {
```
austinmilt, line 9: good opportunity to switch to our new pattern

```suggestion
export function StorefrontProvider(props: {children: React.ReactElement}): JSX.Element {
```

src/modules/storefront/provider.tsx
```diff
@@ -39,11 +31,11 @@ export const StorefrontProvider = ({ wallet, children }: StorefrontProviderProps
       setStorefront(storefront);
       setSearching(false);
     });
-  }, [wallet?.pubkey]);
+  }, [pubkey]);
 
   return (
     <StorefrontContext.Provider value={{ searching, storefront }}>
-      {children({ searching, storefront })}
+      {children}
```
austinmilt, line 38: If taking above suggestion, then:

```suggestion
      {props.children}
```

src/modules/utils/image.ts
```diff
@@ -29,19 +26,16 @@ function tryGetImageFromPublicKey(publicKey?: string | PublicKey | null): string
   return result;
 }
 
-
 export function getFallbackImage(seed?: number | string | PublicKey | null): string {
   let seedNumber: number = 1;
-  if ((seed !== null) && (seed != undefined)) {
+  if (seed !== null && seed != undefined) {
```
austinmilt, line 31: :(

src/views/home/Footer.tsx
```diff
@@ -55,11 +55,11 @@ const Footer: VFC = () => {
               <li>
                 <InternalLink href="/nfts/new">Mint NFTs</InternalLink>
               </li>
-              <li>
+              {/* <li>
```
austinmilt, line 58: intentional?

austinmilt (changes requested):
Mostly small stuff. I do think missing environment variables should throw an error

src/views/home/Footer.tsx
kespinola: Yeah this isn't active yet so good to hide.

kespinola (approved):
LG2M. Largest impact I see is on storefront create and edit if you want to give that a test run. Thank you. @austinmilt has good hygiene comments.

src/lib/utils/index.ts
kristianeboe: This is some crossmint stuff Milo made, he can fix it later :)

mpwsh: my fault here lmao, sorry. it was more like a debugging thing :D

PR #598, opened 2022-07-15
Send price in lamports to sdk transactions

alchemistgo87:


src/lib/utils/conversion.ts
```diff
@@ -0,0 +1,5 @@
+import { LAMPORTS_PER_SOL } from '@solana/web3.js';
```
kespinola, line 1: Can we move this to `src/modules/sol` can use that module to house all sol related utility functions. I want to stay way from a general lib utils blackhole.

kespinola (changes requested):
Looks great just the recommendation on the home for the helper function.

src/lib/utils/conversion.ts
alchemistgo87: Sure!

PR #603, opened 2022-07-15
feat: moved pages to src folder and moved to relative imports everywhere

kristianeboe:
Purely a cosmetic PR that moves the /pages folder into the /src folder and updates almost all files to use relative file paths powered by tsconfig

Also adds a little util as "yarn ts:lint" to display all current build errors in the terminal at once, instead of having to tediously run yarn build over and over. (`npx tsc --noEmit` under the hood)


kristianeboe:
Top level files like sign-metadata are untouched (besides a prettier run)

PR #609, opened 2022-07-18
change address to id for featuredListings query

imabdulbasit:


src/graphql/home/featuredBuyNowListings.graphql
```diff
@@ -1,6 +1,6 @@
 query featuredBuyNowListings($marketplace: String!, $limit: Int!) {
```
austinmilt, line 1: This file should be safe to delete. The query has been moved to `home.graphql`

austinmilt (changes requested):

austinmilt (approved):

PR #641, opened 2022-07-22
UI fixes

kristianeboe:
fixes
- text overflow in Hero feed
- z index issues with popover in alpha (and other places)
- lack of images in popover profile cards
image

image

image

### DIFF
--- a/src/views/alpha/FeedCard.tsx
+++ b/src/views/alpha/FeedCard.tsx
@@ -247,7 +247,15 @@ export const ProfileHandleStack = ({ users }: { users: User[] }) => {
   );
 };
 
-export const ProfileHandle = ({ user, shorten = false }: { user: User; shorten?: boolean }) => {
+export const ProfileHandle = ({
+  user,
+  shorten = false,
+  ...props
+}: {
+  user: User;
+  shorten?: boolean;
+  popoverPlacement?: 'top' | 'bottom';
+}) => {
   const [twitterHandle, setTwitterHandle] = useState(user.profile?.handle);
   const [twitterHandleQuery, twitterHandleQueryContext] = useTwitterHandleFromPubKeyLazyQuery();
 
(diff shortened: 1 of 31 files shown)


src/views/alpha/FeedCard.tsx
```diff
@@ -264,19 +272,24 @@ export const ProfileHandle = ({ user, shorten = false }: { user: User; shorten?:
...
-          address={user.address}
-          context={{
-            data: getProfilePreivewDataFromUser(user),
-            loading: false,
-            refetch: () => {},
-            fetchMore: () => {},
-          }}
-        />
+        <div className="w-80">
+          <ProfilePreview
```
kespinola, line 283:
```suggestion
          <ProfilePreview
            className="w-80"
```

recommend passing the className to the ProfilePreview component. Also noticed the preview could use a little padding at the bottom of the popover.

src/views/alpha/FeedCard.tsx
```diff
@@ -264,19 +272,24 @@ export const ProfileHandle = ({ user, shorten = false }: { user: User; shorten?:
...
+            address={user.address}
+            context={{
+              data: getProfilePreivewDataFromUser(user),
```
kespinola, line 286: NB: Instead of generic utils I think modules/profiles is a good spot for this function. convertToProfile Also instead of converting in the caller consider accepting User and running the conversion within the profile component. Maybe even drop if this is the only spot needing the conversion.

src/views/alpha/FeedCard.tsx
```diff
@@ -247,7 +247,15 @@ export const ProfileHandleStack = ({ users }: { users: User[] }) => {
...
+  user,
+  shorten = false,
+  ...props
+}: {
+  user: User;
+  shorten?: boolean;
+  popoverPlacement?: 'top' | 'bottom';
+}) => {
   const [twitterHandle, setTwitterHandle] = useState(user.profile?.handle);
   const [twitterHandleQuery, twitterHandleQueryContext] = useTwitterHandleFromPubKeyLazyQuery();
```
kespinola, line 261: Why does each feedcard need to fetch the profile? That is a lot of requests. The profile info is available of the feed event query better to consume from there.

kespinola (approved):

src/views/alpha/FeedCard.tsx
kespinola: Also recommend min width here so if the user has collected a lot of NFTs the box stretches past the min width instead of colliding with the created.

src/views/alpha/FeedCard.tsx
kristianeboe: It's a lazy query so it's only called when the required info is not available, but looking into removing it now

src/views/alpha/FeedCard.tsx
kristianeboe: @austinmilt for your consideration

src/views/alpha/FeedCard.tsx
kristianeboe: Noted, will remember this for a later pass of the componenet

src/views/alpha/FeedCard.tsx
austinmilt: > instead of converting in the caller consider accepting User and running the conversion within the profile component

IMO we should be removing Graph QL types from our components completely and convert to the data type needed by the component before passing data through. This helps keep components "dumb" and facilitates refactoring.

PR #656, opened 2022-07-27
Update @solana/wallet-adapter-wallets & add backpack support

echohtp:
New version of `@solana/wallet-adapter-wallets` has added support for Backpack (yet to be released).


milorue (approved):
Awesome, ahead of the curve 😄

Issue #643, opened 2022-07-22
Discovery + Collections Cleanups


bellebellesol:
## **[Discovery](https://www.figma.com/file/tNPfdIHmoevsnKcc8dPMz2/Home-V2?node-id=2353%3A8419)**

- [x]  Remove standalone 'Alpha' link from nav so it only appears once, inside the new discover dropdown menu
- [x]  Tighten up horizontal alignment of icons and their respective labels in dropdown menu. Icons look slightly too raised, should be aligned with the pink lines: 
![Image](https://user-images.githubusercontent.com/95664179/180430205-83447f75-618f-4df9-8823-a42014caf80c.png)

- [x] Implement a few new design changes to tabs/searchbar area (view design in figma). This keeps tabs in the same position regardless of filter panel being expanded or collapsed and feels cleaner, akin to OpenSea and Rarible's explore pages
Filter collapsed:
![Image](https://user-images.githubusercontent.com/95664179/180460804-0be09cb1-0c6f-4820-ad59-727015f41b41.png)
Filter expanded:
![Image](https://user-images.githubusercontent.com/95664179/180460837-25a7d77d-e576-4438-a110-ac72b0c2cd4f.png)

- [x] Tighten up styling around the search bar and sorting dropdown menus. Small deviations from design such as search bar width, expanded dropdown styling etc. 
If its simple and quick, add the active sort item to the top of the dropdown list with a check mark to show its selected currently.

Currently looks like:
![Image](https://user-images.githubusercontent.com/95664179/180434922-041da27c-5f89-4a99-a9df-965a7e199067.png)
![Image](https://user-images.githubusercontent.com/95664179/180434935-b3892d8c-689f-49ad-bb90-3e15156d1147.png)
![Image](https://user-images.githubusercontent.com/95664179/180434936-66f28b62-c078-404a-a8a7-cef644c8d9e0.png)

Should look like:
![Image](https://user-images.githubusercontent.com/95664179/180461226-3c044c27-dbab-47fa-83e2-7599752843cf.png)
![Image](https://user-images.githubusercontent.com/95664179/180435024-f3a18015-9dfb-4a7e-a554-180b48b0bab1.png)
![Image](https://user-images.githubusercontent.com/95664179/180435025-82bba337-4869-443a-bfc0-e962c4eb5ba7.png)

- [x] Adjustments to the layout switcher (pic for context):
[...]

PR #659, opened 2022-07-29
Discovery UI QA

kespinola:
### Changes
- Full width discovery tabs with icons
- Search filters for discovery pages full width.
- Move filters to CardGrid defined within the view instead of the layout.
- Filter sticks under the main filters
- Top marketcap and volume on home page is set to 24 hour period
- Add Select component to replace discovery select filters using headless
- Style active state for search with search icon
- Drop unsupported sorts from nfts page for now as they aren't supported

Screen Shot 2022-08-01 at 9 07 29 PM
Screen Shot 2022-08-01 at 9 17 07 PM
Screen Shot 2022-08-01 at 9 17 21 PM
Screen Shot 2022-08-01 at 9 17 36 PM
Screen Shot 2022-08-01 at 9 18 24 PM

resolves https://github.com/holaplex/holaplex/issues/643


kristianeboe (approved):
Added some tiny tiny tailwind changes, but overall looks great! Only other thing I noticed is the filter icon looks a big huge...

image

src/components/Filters.tsx
```diff
@@ -46,6 +47,18 @@ export default function FiltersSection(props: FiltersSectionProps): JSX.Element
...
 
+interface FilterIconProps {
+  collapsed: boolean;
+  onClick: () => void;
+}
+
+FiltersSection.FilterIcon = function FilterIconButton(props: FilterIconProps) {
+  return (
+    <button onClick={props.onClick} className="rounded-lg hover:border-gray-800">
+      <FilterIcon className="mr-2 w-12 hover:cursor-pointer" />
```
kristianeboe, line 58:
```suggestion
      <FilterIcon className="mr-2 w-8 hover:cursor-pointer" />
```

src/components/CardGrid.tsx
```diff
@@ -68,76 +104,54 @@ enum GridView {
...
+
   /**
    * 4x4 (16 cards visible at a time)
    */
   FOUR_BY_FOUR = '4x4',
   /**
    * 6x6 (36 cards visible at a time)
    */
-  SIX_BY_SIX = '6x6',
+  FIVE_BY_FIVE = '6x6',
```
austinmilt, line 120: Should this be `SIX_BY_SIX` or `'5x5'`?

src/components/CardGrid.tsx
```diff
@@ -68,76 +104,54 @@ enum GridView {
...
 }
 
-const DEFAULT_GRID_VIEW: GridView = GridView.SIX_BY_SIX;
-
 interface GridSelectorProps {
-  onChange: (view: GridView) => void;
+  selected: 0 | 1;
+  onChange: (index: 0 | 1) => void;
```
austinmilt, line 125: Why change from a well-named enum to an ambiguous integer?

src/components/Select.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export function Select<V>({ value, onChange, options }: SelectProps<V>) {
+  const selected = value.label;
+
+  return (
+    <Listbox value={value} onChange={onChange}>
+      {({ open }) => (
+        <div className="relative w-full">
+          <Listbox.Button
+            className={classNames(
```
austinmilt, line 26: `clsx`

src/components/Select.tsx
```diff
@@ -0,0 +1,70 @@
...
+              'relative flex w-full cursor-pointer flex-row items-center justify-between rounded-lg border-2 border-solid border-gray-800 bg-transparent p-2 text-left placeholder-gray-500 shadow-md hover:border-white  focus:placeholder-transparent focus:shadow-none focus:outline-none focus:ring-0',
```
austinmilt, line 27: It would be awesome to break these long class names into more digestible units, taking advantage of `clsx`. I like breaking down more or less by state on the outermost, and by "thing being styled" in inner units, e.g.

```suggestion
              [
                'relative flex w-full', 'flex-row items-center justify-between p-2 text-left', 
                'rounded-lg border border-2 border-solid border-gray-800 bg-transparent shadow-md placeholder-gray-500', 'cursor-pointer'
              ], 
              [
                'hover:border-white  focus:placeholder-transparent focus:shadow-none focus:outline-none focus:ring-0'
              ]
```

austinmilt (changes requested):
Looks to be one typo, but a couple other things to consider as well

src/components/Select.tsx
kristianeboe: Why the [ ] @austinmilt ? Could it not just be separate regular arguments? :) 

```
'relative flex w-full', 'flex-row items-center justify-between p-2 text-left', 
'rounded-lg border border-2 border-solid border-gray-800 bg-transparent shadow-md placeholder-gray-500', 'cursor-pointer',
'hover:border-white  focus:placeholder-transparent focus:shadow-none focus:outline-none focus:ring-0'
```

src/components/CardGrid.tsx
```diff
@@ -25,32 +26,67 @@ export interface CardGridWithHeaderProps<T> {
...
+    return ifElse(
+      isNil,
+      always([GridView.FOUR_BY_FOUR, GridView.FIVE_BY_FIVE]),
+      ifElse(
+        always(collapsed),
+        always([GridView.FOUR_BY_FOUR, GridView.FIVE_BY_FIVE]),
+        always([GridView.THREE_BY_THREE, GridView.FOUR_BY_FOUR])
+      )
+    )(props.filters);
+  }, [props.filters, collapsed]);
```
kespinola, line 53: @kristianeboe I know I promised to not used ramda but its why to productive to not leverage. I'd like to do a workshop or 2 on the topic to kick start folks familiarity with the package.

src/components/Select.tsx
kristianeboe: That being said I think it's more or less fine for this className to be long as it is basically ripped from the styled Tailwind example and well encapsulated into the low level react component itself.

src/components/CardGrid.tsx
kristianeboe: I'm sorry, but I want to push back on that. Even if you could bring our team up to speed on Ramda, you can't hold a workshop for individual contributors who want to engage with and contribute to the open source code base. 

Personally I dont't easily understand what's going on here either, but maybe it's just late in the evening for me.

As for the grid views they might need some rethinking in its entirety, a potentially better solution is to define tailwind grid classes based on type of grid so:
base: grid gap-6
profileGrid: grid-cols-1 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4
nftGrid: grid-cols-2 sm:grid-cols-4 md:grid-cols-6 lg:grid-cols-8
etc.

We can discuss more ofc.

src/components/Select.tsx
austinmilt: The square brackets, like the rest, are just in the name of readability. It's really hard (for me) to read a long class name and get an idea what the element will look like. It's a lot easier for me to read it in units like this:

`'relative flex w-full'`: how is this element positioned in its context
``flex-row items-center justify-between p-2 text-left'`: how does this element position its children
...
`'hover:border-white  focus:placeholder-transparent...'`: how does this element change when in a different state (e.g. hover)

I like to put the square brackets around to separate major categories. Everything in the default state would be wrapped in an outer brackets, and other states in another set of brackets, like `['default state stuff'], ['large screen stuff'], ['hover/focus stuff']`. Since each "state" might have multiple class name strings broken down by, e.g. self position vs child position, I like to group them in the square brackets for that state.

Obviously it's not a strict rule, but it helps me when class names are otherwise very long.

austinmilt: > That being said I think it's more or less fine for this className to be long as it is basically ripped from the styled Tailwind example and well encapsulated into the low level react component itself.

I dont see how this makes a difference. As soon as the code was copy/pasted it became our code, so it should follow our rules

src/components/CardGrid.tsx
kespinola: done switched to GridToggle enum.

src/components/CardGrid.tsx
kespinola: dropped 6 x 6 doesn't work for our cards. also design has 4x4 for compact and 3x3 for expanded.

src/components/Select.tsx
kespinola: ty. done.

src/components/CardGrid.tsx
kespinola: I removed the use of ramda. And we should keep our grid options consistent poor.

src/components/Select.tsx
kespinola: Its overkill to use clsx on every class declaration in order to break up classnames. If folks want to follow a pattern I recommend following a loose ordering where it goes:
- positioning (eg flex, text align, grid)
- box sizing (eg padding, margin)
- styling (eg text color, bg color, border size)
- active states (eg transitions, hover, focus)

When clsx is used for conditionally adding styles I will look to break them up per line. I don't see much benefit from using array notation vs declaring one item per line. This is what I've done here.

bellebellesol:
This is looking great! A few final items I noticed after reviewing on dev:

- [ ] Need to add a hover state for the items under main Discover dropdown. Just apply the same hover state we have for the 2 sorting dropdowns inside the discover page.
Screen Shot 2022-08-02 at 12 12 06 pm

- [ ] Fix skeleton loaders and remove the line of text + spinner that shows up in each tab while the content is loading. Ideally we just have skeletons and if the result doesn't exist we then show the empty state. Seems like skeletons are showing up for NFTs and Collections tabs for a second, but then disappearing, but looks Profiles tab isn't showing any sign of skeletons. 
https://user-images.githubusercontent.com/95664179/182352688-de3c49a8-074b-456d-93a8-172a46a6ed60.mov

- [ ] Fix dropdown being positioned below search/sort bar and check if it's occurring elsewhere / if other scroll positions are working correctly 
Screen Shot 2022-08-02 at 12 16 11 pm

PR #672, opened 2022-08-02
Milo/feat/multi marketplace listings

milorue:
- Added support for other auction house listings directly buying on our platform
- Further discussion needed on how to handle magic eden listings but for now we'll just link our to ME.
- Nft Cards and Details reflect all listings, prioritizing as follows
  - Holaplex listings
  - Cheapest listings
  - everything else

All listings are visible on details page but only the cheapest/holaplex listing is visible on the nft preview card on profiles/collections


src/pages/nfts/[address].tsx
```diff
@@ -205,37 +225,67 @@ export default function NftByAddress({
...
+        return () => clearTimeout(timer);
+      }
+    }, [linkCopied]);
+
+    const handleCopyClick = async () => {
+      await navigator.clipboard.writeText(address || `Error`);
+      setLinkCopied(true);
+    };
+    return (
+      <div className={`flex items-center justify-between`}>
```
Comment (austinmilt) [inline src/pages/nfts/[address].tsx:245]: change template literals to plain strings

src/pages/nfts/[address].tsx
```diff
@@ -420,8 +474,94 @@ export default function NftByAddress({
...
+                          (ah) => ah.address === otherListing?.auctionHouse?.address.toString()
+                        ) ||
+                          MARKETPLACE_PROGRAMS.find(
+                            (ah) => ah.address === otherListing?.marketplaceProgramAddress
+                          ) || {
+                            name: 'Unknown Marketplace',
+                            address: null,
+                            logo: '/images/listings/magiceden.png',
+                            link: 'https://magiceden.io/item-details/',
+                          };
```
Comment (austinmilt) [inline src/pages/nfts/[address].tsx:489]: This looks like a great candidate for a utility function, given where you're using it and how often

src/pages/profiles/[publicKey]/offers.tsx
```diff
@@ -208,15 +208,15 @@ const OfferPage = (props: WalletDependantPageProps) => {
             return nft.offers
               .filter((o) => o.buyer === profilePK)
               ?.sort(byDate)
-              .map((offer) => getActivityCard(nft as Nft, offer as Offer));
+              .map((offer) => getActivityCard(nft as Nft | any, offer as Offer));
```
Comment (austinmilt) [inline src/pages/profiles/[publicKey]/offers.tsx:211]: Is `| any` necessary or is there a more narrowly defined type you could use?

src/views/_global/holaplexConstants.ts
```diff
@@ -1,5 +1,37 @@
 export const HOLAPLEX_MARKETPLACE_SUBDOMAIN = `haus`;
 
+// TODO: deprecate
 export const HOLAPLEX_MARKETPLACE_ADDRESS = `9SvsTjqk3YoicaYnC4VW1f8QAN9ku7QCCk6AyfUdzc9t`;
 
+// TODO: deprecate
+export const OPENSEA_MARKETPLACE_ADDRESS = '3o9d13qUvEuuauhFrVom1vuCzgNsJifeaBYDPquaT73Y';
+
+export const AUCTION_HOUSE_ADDRESSES = [
```
austinmilt, line 9: Can you define an interface for these just so we get some type checking?

austinmilt (changes requested):
a few minor changes. I think, at minimum, it's worth extracting the logic to find auctionHouseInfo into a utility method.

src/pages/profiles/[publicKey]/nfts.tsx
```diff
@@ -34,6 +36,8 @@ import { TailSpin } from 'react-loader-spinner';
 import NoProfileItems, { NoProfileVariant } from '@/components/NoProfileItems';
 import ProfileLayout from '@/views/profiles/ProfileLayout';
 import GridSelector, { GridSize } from '@/components/GridSelector';
+import { AhListingMultiMarketplace } from '../../nfts/[address]';
+import { Tooltip } from 'antd';
```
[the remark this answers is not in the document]
Comment (milorue) [reply src/pages/profiles/[publicKey]/nfts.tsx:40]: it does mixed up the import will update

src/pages/profiles/[publicKey]/offers.tsx
Comment (milorue) [reply src/pages/profiles/[publicKey]/offers.tsx:211]: i have a TODO to add marketplaceProgramAddress to the sdk this is a temp fix until that update

src/pages/profiles/[publicKey]/nfts.tsx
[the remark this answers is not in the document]
Comment (milorue) [reply src/pages/profiles/[publicKey]/nfts.tsx]: https://github.com/holaplex/holaplex/pull/672/commits/e3960d86323770834c6569020163ed5e242a5008

src/views/_global/holaplexConstants.ts
milorue: gotcha did this as part of the utility function work

src/pages/nfts/[address].tsx
Comment (milorue) [reply src/pages/nfts/[address].tsx:246]: good catch an awk habit to break

kespinola (changes requested):
I really think this would be good information to expose through the api.

```graphql
listing {
  auctionHouse {
    address
  }
  marketplace {
    name
     logoUrl
     homepage
  }
}
```

We can maintain a config file on the indexer that gets loaded up at boot then have a dataloader that we can attach to object that support either auction_house or marketplace_program.

Lets discuss at standup and pair during product huddle.

Again we want to expose as much information through the API to provide more value to clients paying for that service. Holaplex.com should be a skin on the API and contain no data. That includes searching, sorts, and auxiliary information like this.

kespinola (commented):
We also may need to strip out receipts when making purchases to other auction houses. I'm reviewing that now.

src/modules/utils/marketplace.ts
```diff
@@ -34,3 +42,17 @@ export const errorCodeHelper = (err: string) => {
...
+      (ah) => ah.address === listing?.auctionHouse?.address.toString()
+    ) ||
+    MARKETPLACE_PROGRAMS.find((ah) => ah.address === listing?.marketplaceProgramAddress) || {
+      name: 'Unknown Marketplace',
+      address: null,
+      logo: '/images/listings/unknown.svg',
+      link: '',
+    }
+  );
+};
```
kespinola, line 58: This is possible a good option for a graphql client side field resolve. A good spot to grab the info until its moved to the API.

https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies

kespinola: Also instead of modules/utils/marketplace this should be in a modules/marketplace. Lets try not to have a utils module it just means we haven't come up with a good grouping name for the set of functions.

src/modules/utils/marketplace.ts
```diff
@@ -1,3 +1,11 @@
+import { AhListing } from '@holaplex/marketplace-js-sdk';
+import { AhListingMultiMarketplace } from '../../pages/nfts/[address]';
+import {
+  AUCTION_HOUSE_ADDRESSES,
+  AUCTION_HOUSE_INFO,
+  MARKETPLACE_PROGRAMS,
+} from '../../views/_global/holaplexConstants';
```
kespinola, line 7: The constants should be in a marketplaces module and not pulled from views. The import direction should be from modules into views but not the reverse.

src/pages/profiles/[publicKey]/offers.tsx
Comment (kespinola) [reply src/pages/profiles/[publicKey]/offers.tsx:211]: you can extend the Nft interface or join to together for this call.

```typescript
type IFooBar = IFoo & IBar;
```

https://stackoverflow.com/questions/49723173/merge-two-interfaces

src/pages/nfts/[address].tsx
```diff
@@ -38,9 +41,17 @@ import { seededRandomBetween } from '@/modules/utils/random';
...
-// import { Tag as FeatherTag, Zap } from 'react-feather';
+import { DollarSign, Tag as FeatherTag, Zap } from 'react-feather';
+import Popover from '../../components/Popover';
 import { LightningBoltIcon, TagIcon } from '@heroicons/react/outline';
+import { getAuctionHouseInfo } from '../../modules/utils/marketplace';
+
+// TODO: update sdk to include marketplaceProgramAddress
+export interface AhListingMultiMarketplace extends AhListing {
+  marketplaceProgramAddress?: string;
+}
```
Comment (kespinola) [inline src/pages/nfts/[address].tsx:54]: It only takes a sec to publish the marketplace-js-sdk. I recommend you update that first then use here.

src/pages/nfts/[address].tsx
```diff
@@ -165,6 +176,16 @@ export default function NftByAddress({
   const defaultListing = nft?.listings.find(
     (listing) => listing?.auctionHouse?.address.toString() === HOLAPLEX_MARKETPLACE_ADDRESS
   );
+
+  const otherListings = nft?.listings.filter(
+    (listing) => listing.auctionHouse?.address.toString() !== HOLAPLEX_MARKETPLACE_ADDRESS
+  );
+
+  const magicEdenListing = nft?.listings.find((listing) => listing.auctionHouse === null);
```
Comment (kespinola) [inline src/pages/nfts/[address].tsx:184]: Should check on the marketplace program instead of auctionHouse being null.

src/pages/nfts/[address].tsx
```diff
@@ -165,6 +176,16 @@ export default function NftByAddress({
...
+  const openSeaListing = nft?.listings.find(
+    (listing) => listing?.auctionHouse?.address.toString() === AUCTION_HOUSE_ADDRESSES[1].address
```
Comment (kespinola) [inline src/pages/nfts/[address].tsx:186]: A magic index is going to get flaky fast. Should assign open sea auction house address to a constant and reference that.

src/pages/nfts/[address].tsx
Comment (milorue) [reply src/pages/nfts/[address].tsx:184]: oh woops i no longer use this in favor of the dynamic way

PR #673, opened 2022-08-04
Discover Card Grid Adjustments

kespinola:
### Changes
- Add preview state for profiles
- Instead of has more spinner show a row of preview cards
- Stop show the "no data" fallback text for a moment while the page is initializing
- Dismiss the has more when no more results returned from the api

Screen Shot 2022-08-04 at 3 08 46 PM
Screen Shot 2022-08-04 at 3 13 40 PM
Screen Shot 2022-08-04 at 3 14 43 PM
Screen Shot 2022-08-04 at 3 14 48 PM
Screen Shot 2022-08-04 at 3 14 54 PM

### DIFF
--- a/src/components/CardGrid.tsx
+++ b/src/components/CardGrid.tsx
@@ -2,10 +2,9 @@ import { DoubleGrid } from '@/assets/icons/DoubleGrid';
 import { TripleGrid } from '@/assets/icons/TripleGrid';
 import FiltersSection, { FilterProps } from '@/components/Filters';
 import clsx from 'clsx';
-import { useState, useEffect, useMemo } from 'react';
+import { useState, useMemo } from 'react';
 import { DebounceInput } from 'react-debounce-input';
 import { InView } from 'react-intersection-observer';
-import { TailSpin } from 'react-loader-spinner';
 import { SearchIcon } from '@heroicons/react/outline';
 
 export interface CardGridWithHeaderProps<T, F> {
@@ -296,9 +295,6 @@ export interface CardGridProps<T> {
  * @returns
  */
 export function CardGrid<T>(props: CardGridProps<T>): JSX.Element {
-  const [bodyElements, setBodyElements] = useState<JSX.Element[]>([
-    props.cardContext.noDataFallback ?? <></>,
-  ]);
   const gridId: string = useMemo(() => `grid-${Math.round(Math.random() * 100000)}`, []);
 
   let gridViewClasses: string;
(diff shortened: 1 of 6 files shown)


src/components/CardGrid.tsx
```diff
@@ -328,53 +324,54 @@ export function CardGrid<T>(props: CardGridProps<T>): JSX.Element {
...
         ))}
-      </div>
-      <InView threshold={0.1} onChange={props.dataContext.onLoadMore} className="w-full">
-        <div
-          className={clsx('my-6 flex w-full items-center justify-center font-bold', {
-            hidden:
-              !props.dataContext.hasMore ||
-              (props.dataContext.data && props.dataContext.data.length === 0),
+        {props.dataContext.loading &&
+          [...Array(gridCols * gridCols).keys()].map((i) => {
```
austinmilt, line 348:
```suggestion
          [...Array(gridCols * gridRows).keys()].map((i) => {
```

src/views/discover/discover.hooks.ts
```diff
@@ -161,12 +170,22 @@ export function useDiscoverCollectionsByVolumeLazyQueryWithTransforms(
...
+    (previous, more, setHasMore) => {
       if (!more) return previous;
+
+      if (setHasMore) {
+        if (more.collectionsFeaturedByVolume.length === 0) {
+          setHasMore(false);
+        } else {
+          setHasMore(true);
+        }
+      }
```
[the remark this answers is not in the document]
austinmilt: This should already be taken care of by `useHolaplexInfiniteScrollQuery`. Sorry, should have added some docs to that, but it's main purpose is to take care of lazy loading more rows. It returns a stateful variable `hasMore` that gets updated when the lazy loading is triggered.

austinmilt (changes requested):
Shouldnt need the custom hasMore logic.

src/views/discover/discover.hooks.ts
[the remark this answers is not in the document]
kespinola: it wasn't handling it correctly.

src/components/CardGrid.tsx
kespinola: There is no grid rows. Are you suggesting that is added?

src/hooks/useApolloQuery.ts
```diff
@@ -69,7 +73,6 @@ export function useHolaplexInfiniteScrollQuery<
   const onCompleted: (raw: TGraphQLDataObject) => void = useCallback(
     (raw) => {
       const rawDataArray: TGraphQLElement[] | undefined = listExtractor(raw);
-      setHasMore(rawDataArray != null && rawDataArray.length > 0);
```
kespinola, line 72: This was the previous location looking to handle the has more but once there is data returned this is always greater than 0.

src/components/CardGrid.tsx
austinmilt: Ah, no. I was thinking of this as a fixed-size grid when I saw that

src/views/discover/discover.hooks.ts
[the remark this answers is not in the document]
austinmilt: Ah, great catch! I would love if, rather than try to put in a bandaid here, we could make `useHolaplexInfiniteScrollQuery` work correctly. I'm happy to take a stab at it

austinmilt (approved):

PR #677, opened 2022-08-08
Milo/feat/cookies banner

milorue:
image

image

- Added cookie banner & settings modal
- Added a cookie settings link to the footer

### DIFF
--- a/src/components/AppHeader.tsx
+++ b/src/components/AppHeader.tsx
@@ -65,7 +65,8 @@ export function AppHeader() {
         <div className="flex w-full flex-row items-center py-1 text-2xl font-bold">
           <Link href="/" passHref>
             <a className="flex justify-between font-bold">
-              👋&nbsp;&nbsp;<span className="hidden md:inline-block">Holaplex</span>
+              <span className={'animate-waving'}>👋</span>&nbsp;&nbsp;
+              <span className="hidden md:inline-block">Holaplex</span>
             </a>
           </Link>
           {/* TODO: temp disabled for deploy */}
--- a/src/components/CookieBanner.tsx
+++ b/src/components/CookieBanner.tsx
@@ -1,52 +1,116 @@
 import React, { useState } from 'react';
 import { useCookies } from 'react-cookie';
-import ReactDom from 'react-dom';
 import Button from './Button';
+import Modal from './Modal';
+import { Switch } from '@headlessui/react';
+import { LockClosedIcon } from '@heroicons/react/outline';
+import CookieSettings from './CookieSettings';
 
 export const COOKIES_ACCEPTED = 'holaplex_cookies_acceptance';
+export const ANALYTICS_ACCEPTED = 'holaplex_analytics_cookies_acceptance';
+export const PREFERENCE_ACCEPTED = 'holaplex_preference_cookies_acceptance';
 
-const CookieBanner = () => {
-  const [cookies, setCookie, removeCookie] = useCookies([COOKIES_ACCEPTED]);
+export function useCookieSettings() {
+  const [cookies, setCookie, removeCookie] = useCookies([
+    COOKIES_ACCEPTED,
+    ANALYTICS_ACCEPTED,
+    PREFERENCE_ACCEPTED,
+  ]);
+  const [openSettings, setOpenSettings] = useState<boolean>(false);
   const [tempDecline, setTempDecline] = useState<boolean>(false);
 
-  if (cookies[COOKIES_ACCEPTED] === 'true' || tempDecline) {
+  const setCookieSettings = ({
+    analyticsCookies,
+    essentialCookies = true,
+    preferenceCookies,
+  }: {
+    analyticsCookies: boolean;
+    essentialCookies: boolean;
+    preferenceCookies: boolean;
+  }) => {
+    setCookie('holaplex_analytics_cookies_acceptance', analyticsCookies);
+    setCookie('holaplex_cookies_acceptance', essentialCookies);
+    setCookie('holaplex_preference_cookies_acceptance', preferenceCookies);
+    setOpenSettings(false);
+  };
+
+  const acceptAll = () => {
+    setCookie('holaplex_analytics_cookies_acceptance', true);
+    setCookie('holaplex_cookies_acceptance', true);
+    setCookie('holaplex_preference_cookies_acceptance', true);
+    setOpenSettings(false);
+  };
+
+  const declineCookies = () => {
+    setCookie('holaplex_analytics_cookies_acceptance', false);
+    setCookie('holaplex_cookies_acceptance', false);
+    setCookie('holaplex_preference_cookies_acceptance', false);
+    setTempDecline(true);
+    setOpenSettings(false);
+  };
+
+  return {
+    cookies,
+    setCookie,
+    removeCookie,
+    setCookieSettings,
+    openSettings: setOpenSettings,
+    isSettingsOpen: openSettings,
+    isDeclined: tempDecline,
+    declineCookies,
+    acceptAllCookies: acceptAll,
+  };
+}
+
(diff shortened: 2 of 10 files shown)


src/components/AppHeader.tsx
```diff
@@ -65,7 +65,8 @@ export function AppHeader() {
         <div className="flex w-full flex-row items-center py-1 text-2xl font-bold">
           <Link href="/" passHref>
             <a className="flex justify-between font-bold">
-              👋&nbsp;&nbsp;<span className="hidden md:inline-block">Holaplex</span>
+              <span className={'animate-waving'}>👋</span>&nbsp;&nbsp;
```
austinmilt, line 68: unnecessary curly's

src/components/CookieBanner.tsx
```diff
@@ -0,0 +1,118 @@
...
+    essentialCookies = true,
+    preferenceCookies,
+  }: {
+    analyticsCookies: boolean;
+    essentialCookies: boolean;
+    preferenceCookies: boolean;
+  }) => {
+    setCookie('holaplex_analytics_cookies_acceptance', analyticsCookies);
+    setCookie('holaplex_cookies_acceptance', essentialCookies);
+    setCookie('holaplex_preference_cookies_acceptance', preferenceCookies);
```
austinmilt, line 33: Can you use your consts in here?

src/components/CookieBanner.tsx
```diff
@@ -0,0 +1,118 @@
...
+    setCookieSettings,
+    openSettings: setOpenSettings,
+    isSettingsOpen: openSettings,
+    isDeclined: tempDecline,
+    declineCookies,
+    acceptAllCookies: acceptAll,
+  };
+}
+
+const CookieBanner = () => {
```
austinmilt, line 65:
```suggestion
function CookieBanner(): JSX.Element {
```

src/components/CookieBanner.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const { cookies, openSettings, isSettingsOpen, isDeclined, declineCookies, acceptAllCookies } =
+    useCookieSettings();
+
+  if (cookies[COOKIES_ACCEPTED] === 'true' || isDeclined || typeof window === undefined) {
+    return null;
+  } else {
+    return (
+      <>
+        <div className={`fixed bottom-5 z-50 flex w-full justify-center`}>
```
austinmilt, line 74: change to regular strings and remove unnecessary curly's

src/components/CookieBanner.tsx
```diff
@@ -0,0 +1,118 @@
...
+import Modal from './Modal';
+import { Switch } from '@headlessui/react';
+import { LockClosedIcon } from '@heroicons/react/outline';
+import CookieSettings from './CookieSettings';
+
+export const COOKIES_ACCEPTED = 'holaplex_cookies_acceptance';
+export const ANALYTICS_ACCEPTED = 'holaplex_analytics_cookies_acceptance';
+export const PREFERENCE_ACCEPTED = 'holaplex_preference_cookies_acceptance';
+
+export function useCookieSettings() {
```
austinmilt, line 13: Can you put a return type on this?

austinmilt, line 13: I think we agreed to put exported hooks in to our hooks dir

src/components/CookieSettings.tsx
```diff
@@ -0,0 +1,135 @@
...
+          >
+            Learn more
+          </a>
+        </p>
+        <form onSubmit={saveSettings} className={`flex flex-col gap-4`}>
+          <div className={'flex items-center justify-between gap-2'}>
+            <div className={'flex flex-col gap-2'}>
+              <h6 className={'m-0 text-base'}>Essential</h6>
+              <p className={'m-0 text-xs text-gray-300'}>
+                {"Necessary for the website to function properly and can't be deactivated"}
```
austinmilt, line 65: As a site user, I would want to see what these essential cookies are to decide whether or not I actually want to use the site. Otherwise, the "Essential" is relatively meaningless.

src/lib/utils/index.ts
```diff
@@ -73,3 +73,19 @@ export const RUST_ISO_UTC_DATE_FORMAT = 'yyyy-MM-dd HH:mm:ss';
 export function isTouchScreenOnly(): boolean {
   return window.matchMedia('(hover: none) and (pointer: coarse)').matches;
 }
+
+export function getCookie(cname: string) {
```
austinmilt, line 77: this also looks like it should go into another directory... perhaps the views

austinmilt, line 77: nit: `cname` is a bit confusing (compare with e.g. cname lookup), and not really more helpful than just calling this `name`

src/lib/utils/index.ts
```diff
@@ -73,3 +73,19 @@ export const RUST_ISO_UTC_DATE_FORMAT = 'yyyy-MM-dd HH:mm:ss';
...
+  let decodedCookie = decodeURIComponent(document.cookie);
+  let ca = decodedCookie.split(';');
+  for (let i = 0; i < ca.length; i++) {
+    let c = ca[i];
+    while (c.charAt(0) == ' ') {
+      c = c.substring(1);
+    }
+    if (c.indexOf(name) == 0) {
+      return c.substring(name.length, c.length);
+    }
```
austinmilt, line 88: could this be simplified to
```suggestion
    let c = ca[i].trim();
    if (c.startsWith(name)) {
      return c.substring(name.length, c.length);
    }
```

I think you could further simplify the function by not appending `=` to the name and then just doing

```
return c.substring(name.length + 1, c.length);
```

src/lib/utils/index.ts
```diff
@@ -73,3 +73,19 @@ export const RUST_ISO_UTC_DATE_FORMAT = 'yyyy-MM-dd HH:mm:ss';
...
+  }
+  return '';
```
austinmilt, line 90: Dont love this fallback, as it implies (to me) that the cookie exists, but its value has been set to `''`. IMO better to either throw an error or return `undefined` (or `null`)

src/components/CookieBanner.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useCookies } from 'react-cookie';
+import Button from './Button';
+import Modal from './Modal';
+import { Switch } from '@headlessui/react';
+import { LockClosedIcon } from '@heroicons/react/outline';
+import CookieSettings from './CookieSettings';
+
+export const COOKIES_ACCEPTED = 'holaplex_cookies_acceptance';
+export const ANALYTICS_ACCEPTED = 'holaplex_analytics_cookies_acceptance';
+export const PREFERENCE_ACCEPTED = 'holaplex_preference_cookies_acceptance';
```
austinmilt, line 11: should these be exported from e.g. the views dir instead of in the component (thinking about the direction of dependencies)?

austinmilt (changes requested):

src/components/CookieBanner.tsx
milorue: Yep auto complete made me do it 😊

src/lib/utils/index.ts
milorue: Agreed

src/components/CookieBanner.tsx
milorue: I think they're fine as part of the component for now if it becomes an issue i'll move them, keeping cookie related stuff together seems like the best pattern atm

src/lib/utils/index.ts
milorue: agreed i think just returning undefined should have the desired effect

src/lib/utils/index.ts
milorue: good suggestion gonna simplify

